PR #5, opened 2017-03-21
Distribution stats and improvements

bartekn:
* Added distribution stats that can also be accessed via a simple API.
* Created a generic `AmountWidget` to display amounts. It's used as a parent class in Total Lumens widget to display total number of lumens but also in Friendbot to display current balance of friendbot.
* Added API link in top right corner of widgets if data can be accessed via API or Horizon.
* Made builds smaller by loading required `lodash` and `react-d3-components ` functions instead of full library code.
* Added smaller improvements and removed unused code.

### DIFF
--- a/app.js
+++ b/app.js
@@ -1,11 +1,18 @@
+// Needed to load lumens.js. Will switch to full ES6 support in backend
+// app soon.
+require("babel-register")({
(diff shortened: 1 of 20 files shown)


jedmccaleb:
r+

PR #10, opened 2017-09-19
WIP: More metrics

poliha:
This is a work in progress for adding data  about
- DEX volume over time
- No. of operations over time:
- No. of anchors over time
- No. of assets over time
- No. of accounts over time
This needs to be running for a few days before we can see the charts take shape.
 @bartekn please glance through and let me know if I'm headed in the right direction.

### DIFF
--- a/app.js
+++ b/app.js
@@ -400,7 +400,7 @@ function getDexVolume() {
     });
   });
 }
-setInterval(getDexVolume(), 3600000);
+setInterval(getDexVolume, 3600000);
 getDexVolume();
 
 // Stream ledgers - get last paging_token
--- a/app.js
+++ b/app.js
@@ -21,7 +21,6 @@ const NODE_TIMEOUT = -2;
 const MONITOR_URL = 'http://core-mon-001.prd.stellar001.external.stellar-ops.com:8001/';
 
 var app = express();
-
 app.set('port', (process.env.PORT || 5000));
 app.set('json spaces', 2);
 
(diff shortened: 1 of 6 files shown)


app.js
```diff
@@ -120,6 +123,96 @@ app.get('/api/nodes', function(req, res) {
...
+    day = day.subtract(1, 'days')
+  }
+
+  multi.exec(function (err, redisRes) {
+    if (err) {
+      res.sendStatus(500);
+      console.error(err);
+      res.error("Error");
+      return;
+    }
```
bartekn: We use very similar code in 3 places now. Maybe let's extract it to a new function?

poliha: yep I extracted this.

poliha:
So I think this is getting close to being ready, just need some input of the arrangements of the charts. 
@jedmccaleb any input?

app.js
```diff
@@ -231,6 +315,77 @@ function checkNodes() {
...
+        console.error(err);
+        return;
+      }
+      console.log("Added DEX volume stats: "+key+" "+res);
+    });
+  });
+}
+setInterval(getDexVolume, 3600000);
+getDexVolume();
+
```
bartekn: A lot of code is duplicated here so let's find a better design. Here's my idea how to solve it:
* Let's move all functions that get data from external services to another package (`data-providers`?). Let's call them `DataProviders`. So we will have: `AccountsDataProvider`, `AssetDataProvider`, `DEXDataProvider`.
* `Storage` class will be responsible for saving and loading data from Redis (gets `redis` object in contructor). It will have the following methods:
  * `updateAccountsCount(date, count)`
  * `updateAssetsCount(date, count)`
  * `updateOperationsCount(date, count)`
  * `updateTransactionsCount(date, count)`
  * `updateDEXVolume(date, asset, volume)`
  * ...other `update` methods.
  * `getAccountsCount(days)` - return number of accounts in the last `days` days
  * `getAssetsCount(days)`
  * `getOperationsCount(days)`
  * `getDEXVolume(days, asset)`
  * ...other `get` methods.
* Finally we will have `Collector` class that will import all data providers and call them periodically to get data. It will also accept `Storage` object in contructor to persist data.
* Let's move all controllers functions (`function(req, res)`) to `controllers` package. Controllers should use `Storage` to get data from Redis.

Let me know if you have questions or other idea how to design it.

app.js
```diff
@@ -231,6 +315,77 @@ function checkNodes() {
...
+    multi.exec(function (err, res) {
+      if (err) {
+        console.error(err);
+        return;
+      }
+      console.log("Added DEX volume stats: "+key+" "+res);
+    });
+  });
+}
+setInterval(getDexVolume, 3600000);
```
bartekn: I think we should call it more often. If it gets called at `11:01pm` we will lose information about all accounts created/deleted in the last hour of that day. I think calling this every 5 minutes or even every minute shouldn't be a problem for a `core-mon` database.

src/components/App.js
```diff
@@ -220,6 +224,23 @@ export default class App extends React.Component {
...
+            <TradeVolumeChart assetPair="XLM_JPY" />
+            <TradeVolumeChart assetPair="XLM_PHP" />
+            <TradeVolumeChart assetPair="XLM_XRP" />
+            <TradeVolumeChart assetPair="XLM_GFT" />
+            <TradeVolumeChart assetPair="XLM_SEED" />
+            <TradeVolumeChart assetPair="XLM_XEL" />
+            <TradeVolumeChart assetPair="XLM_LTC" />
+            <TradeVolumeChart assetPair="XLM_EQD" />
+            <TradeVolumeChart assetPair="XLM_ICN" />
+          </section>
```
bartekn: It's missing a margin:
screen shot 2017-09-20 at 21 12 21

src/components/OperationsChart.js
```diff
@@ -0,0 +1,71 @@
...
+        this.setState({loading: false, data});
+      });
+  }
+
+  render() {
+    return (
+      <div ref={(el) => { this.panel = el; }}>
+        <Panel>
+          <div className="widget-name">
+            <span style={{borderBottom: '2px solid #0074B7'}}>
```
bartekn: We don't need underline if there is only one data group in a chart.

src/components/OperationsChart.js
```diff
@@ -0,0 +1,71 @@
+import React from 'react';
+import Panel from 'muicss/lib/react/panel';
+import axios from 'axios';
+import {scale, format} from 'd3';
+import LineChart from 'react-d3-components/lib/LineChart';
+import clone from 'lodash/clone';
+import each from 'lodash/each';
+
+export default class OperationsChart extends React.Component {
```
bartekn: Isn't it the same data we present in `PublicNetworkLedgersHistoryChart`?

bartekn:
What's the status of this PR?

poliha:
This was moved to internal. Think this can be closed

PR #15, opened 2017-12-14
Update nodes.js

gavinito:


common/nodes.js
```diff
@@ -113,11 +113,19 @@ module.exports = [
...
+    name: "cryptomover host 1",
     host: "cryptodealer.hk",
     port: 11625,
     publicKey: "GCGWABAQ6OUOVUGWJVPRJ5LWBIWYN3CVOVOZYBNQQGIBRULQHYNGQ7GH"
   },
+  {
+    id: "cryptomover",
+    verified: true,
+    name: "cryptomover host 2",
+    host: "cryptomover.org:8000",
```
bartekn: You should remove port from here.

gavinito:
We can't have the JSON serve on port 8000 instead of 80?

PR #16, opened 2017-12-15
Update nodes.js

gavinito:
Added second Cryptomover full node.


bartekn:
Are you sure this node is running? I can't connect to it.

PR #26, opened 2018-02-07
Update nodes.js

n-vishnu:


bartekn:
We'll be redesigning Nodes section this and next week. I'll merge this after redesign is finished.

PR #31, opened 2018-02-26
Redesigned nodes for simpler view

robertDurst:
*NOTE:* Small view is at 1150 px instead of 768px. Was having issues with responsive sizing and text overlap with public key at smaller values. Also, code may be a little messy -- had some troubles making the table responsive without overlapping text. If there are certain parts to fix, please let me know.

## Full Screen

screen shot 2018-02-25 at 8 02 13 pm

## Full Screen Hover

ezgif com-video-to-gif

## Small Screen

screen shot 2018-02-25 at 8 03 34 pm

## Small Screen Expanded View

screen shot 2018-02-25 at 8 03 42 pm

### DIFF
--- a/src/components/Node.js
+++ b/src/components/Node.js
@@ -103,34 +103,36 @@ export default class Node extends React.Component {
     return (
       <div ref={(el) => { this.panel = el; }}>
         <Panel className="mui-col-md-12 node-panel">
-          <div className="mui-col-md-3 mui-col-xs-4 node-circle-container">
-            <div className={this.renderColor()} />
-              {this.props.data.name}
-            </div>
-            <div className={"mui-col-md-3 mui-col-xs-4"}>
-              {this.props.data.host}:{this.props.data.port}
-            </div>
-            {this.state.loading ?
-              'Loading...' :
-              <div className={"mui-col-md-6 mui-col-xs-4 node-barchart-container"}>
-                {
-                  this.state.small ? "" :
-                  <div className="node-hovered-container">
-                    <p className={"node-pubkey"}>{this.props.data.publicKey} </p>
-                  </div>
-                }
-                <div className={this.state.small ? "" : "node-barchart"}>
-                  <BarChart
-                    data={this.renderData()}
-                    width={this.state.chartWidth}
-                    colorScale={this.colorScale}
-                    height={this.state.chartHeigth} />
-                </div>
-                {
-                  this.state.small ? <i className="material-icons node-dropdown-button" onClick={() => this.setState({expanded : !this.state.expanded})}>expand_more</i> : ""
-                }
+          <div className="node-panel-large">
+            <div className="mui-col-md-3 mui-col-xs-4 node-circle-container">
+              <div className={this.renderColor()} />
+                {this.props.data.name}
               </div>
-            }
+              <div className={"mui-col-md-3 mui-col-xs-4"}>
+                {this.props.data.host}:{this.props.data.port}
+              </div>
+              {this.state.loading ?
+                'Loading...' :
+                <div className={"mui-col-md-6 mui-col-xs-4 node-barchart-container"}>
+                  {
+                    this.state.small ? "" :
+                    <div className="node-hovered-container">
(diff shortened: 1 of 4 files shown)


robertDurst:
Added fixes. Still extra whitespace -- although it is much less significant with width < 1200px.

src/components/Nodes.js
```diff
@@ -16,19 +16,17 @@ export default class Nodes extends React.Component {
   }
 
   getUptimeData() {
-    axios.get('/api/nodes')
+    axios.get('http://localhost:3010/data')
```
bartekn: We have a special `DEV` mode you can run by starting `gulp` in one terminal window and `DEV=true node app.js` in the other. This mode creates a proxy to `browser-sync` server so you can use backend API without a need to care about CORS etc.

I also improved this mode to generate random measurements for each node. Please rebase master to use this new feature. Ping me in case of problems.

src/components/Node.js
```diff
@@ -7,15 +7,31 @@ export default class Node extends React.Component {
...
+        this.setState({small: true});
+      } else if (window.innerWidth >= 1150 && this.state.small) {
+        this.setState({small: false, expanded : false});
+      }
+      if (this.state.chartWidth != value/3 && !this.state.small ||
+          this.state.chartWidth != value/20 && this.state.small ) {
+        if (this.state.small) {
+          this.setState({chartWidth: value/20});
+        } else {
+          this.setState({chartWidth: value/3});
```
bartekn: Currently bars in the chars are too thick. They should be the same as in other charts like ledger close time and tx&ops charts. To achieve this let's make the chart wider (`value/2`) and change backend API to return 150 measurements instead of 70 (one-line change in `app.js`).

src/components/Node.js
```diff
@@ -69,60 +85,94 @@ export default class Node extends React.Component {
...
-                    this.state.state == 'up' ?
-                      <strong style={{color: "#2196f3"}}>Up!</strong>
-                      :
-                    this.state.state == 'problems' ?
-                      <strong style={{color: "orange"}}>Problems...</strong>
-                      :
-                      <strong style={{color: "red"}}>Down!</strong>
-                    :
-                    <span>Loading...</span>
+                  this.state.small ? <i className="material-icons" onClick={() => this.setState({expanded : !this.state.expanded})}>expand_more</i> : ""
```
bartekn: Add `cursor: pointer` to CSS of an arrow.

src/scss/_main.scss
```diff
@@ -241,3 +241,98 @@ section {
...
+  align-items: center;
+  visibility: hidden;
+  opacity: 0;
+  z-index: 10;
+  background-color: white;
+  height: 25px;
+  right: 100px;
+  padding: 5px;
+  position: absolute;
+  transition: all ease-out 1s;
```
bartekn: Let's change every animation speed here to something faster. `0.1s` should be good.

src/scss/_main.scss
```diff
@@ -241,3 +241,101 @@ section {
...
+  display: flex;
+  justify-content:flex-end;
+  align-items: center;
+  position: relative;
+}
+
+.node-barchart-container .domain {
+  display: none;
+  visibility: hidden;
+}
```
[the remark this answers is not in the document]
robertDurst: Yes, it took away the graph around the barchart. Here is what it looks like removed

screen shot 2018-03-01 at 2 14 35 pm

src/scss/_main.scss
```diff
@@ -241,3 +241,101 @@ section {
...
+  position: absolute;
+  left: 0;
+  transition: all ease-out 100ms;
+}
+
+.node-pubkey {
+  text-align: center;
+  background-color: white;
+  width: 85%;
+  padding-top: 8px;
```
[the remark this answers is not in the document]
robertDurst: Fixed the vertical alignment issue, there were a few things causing that. Having a challenege reproducing the css issue above.

src/scss/_main.scss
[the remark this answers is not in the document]
bartekn: OK, it's `BarChart`'s class. 👍

PR #47, opened 2018-06-25
Added Collegecoin validator

macgyver2028:
Added Collegecoin US Validator.


MonsieurNicolas:
if this is still a valid request, please rebase it

PR #52, opened 2018-08-13
Add Smartlands validator

MikhailTamarin:


common/nodes.js
```diff
@@ -265,6 +265,13 @@ module.exports = [
     port: 11625,
     publicKey: "GCWJKM4EGTGJUVSWUJDPCQEOEP5LHSOFKSA4HALBTOO4T4H3HCHOM6UX"
   },
+  {
+    id: "smartlands",
+    name: "Smartlands Platform",
+    host: "stellar.smartlands.io",
+    port: 11625,
+    publicKey: "GCBAQCWXKIIOQOEDIW6EB7GY6UREHPTKK26NZFLVK63ZTRFSXDVX2AHV"
```
bartekn: Make sure the public key is visible in smartlands.io [stellar.toml](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0001.md) file in `OUR_VALIDATORS` section.

common/nodes.js
```diff
@@ -265,6 +265,13 @@ module.exports = [
...
+  },
```
bartekn: Add this as the last entry in this file, with the `,` appended.

common/nodes.js
MikhailTamarin: @bartekn added public key to OUR_VALIDATORS section at https://smartlands.io/.well-known/stellar.toml

common/nodes.js
MikhailTamarin: @bartekn added this as the last entry in this file, with the , appended.

PR #58, opened 2018-10-02
Update nodes.js

Hayarobee:
Add Sonic Wallet Validator


vcarl:
Merged #55, so this is redundant

PR #70, opened 2019-01-09
Add banner for testnet reset in January 2019.

theaeolianmachine:
Tested this locally on my machine, and the link resolved and it looked good. Let me know if you need anything else from me.


theaeolianmachine:
@bartekn (since I can't assign or set you as a reviewer).

PR #78, opened 2019-02-05
Remove republic.co

bartekn:
Has been down for more that 30 days.


theaeolianmachine:
@bartekn unfortunately I don't have write access to this repo; I don't think it's part of the platform team's repos (perhaps it should be)?

bartekn:
> I don't think it's part of the platform team's repos (perhaps it should be)?

Fixed.

PR #82, opened 2019-02-22
Add FeeStats widget

bartekn:
Simple widget displaying `/fee_stats` endpoint data:

screen shot 2019-02-22 at 18 58 40


bartekn:
Yeah, this is a temporary solution to display this data fast, will be improved in a future.

Issue #107, opened 2019-05-29
remove node list from the dashboard


tomerweller:
- The node list is incomplete 
- It requires manual updates
- It's misleading as people might think that the availability metric relates to validation while in fact it's just overlay (https://github.com/stellar/dashboard/issues/49)
- removing it allows the dashboard to be focused on one thing: ledger activity.    
- stellarbeat.io does a better job. we can link to that

kissingyou1:
cryptomover.net:11625

PR #116, opened 2019-08-09
Remove smoove.net nodes

chriserice:
Decomission smoove.net nodes


vcarl:
We intend to remove nodes from the dashboard (#107) in favor of [Stellarbeat's node list](https://stellarbeat.io/nodes), so I'm going to close this

PR #127, opened 2019-11-13
Restore and update API checks for lumen counts

piyalbasu:
WIP: DO NOT MERGE

### DIFF
--- a/backend/lumens.js
+++ b/backend/lumens.js
@@ -1,54 +1,51 @@
 import * as commonLumens from "../common/lumens.js";
 
-let cachedData = {
-  "updatedAt": "2019-11-04T19:03:21.898Z",
-  "totalCoins": "105443902087.3472865",
-  "availableCoins": "20054779553.6822807",
-  "distributedCoins": "8772645809.0519286",
-  "programs": {
-    "directProgram": "5435066959.8964838",
-    "bitcoinProgram": "2037756769.6575473",
-    "partnershipProgram": "1138014496.4978975",
-    "buildChallenge": "161807583"
-  }
-};
+let cachedData;
 
 export const handler = function(req, res) {
   res.send(cachedData);
 };
 
-// function updateApiLumens() {
-//   Promise.all([
-//     commonLumens.totalCoins("https://horizon.stellar.org"),
-//     commonLumens.availableCoins(),
-//     commonLumens.distributionAll(),
-//     commonLumens.distributionDirectSignup(),
-//     commonLumens.distributionBitcoinProgram(),
-//     commonLumens.distributionPartnershipProgram(),
-//     commonLumens.distributionBuildChallenge(),
-//   ])
-//     .then(function(data) {
-//       var response = {
-//         updatedAt: new Date(),
-//         totalCoins: data[0],
-//         availableCoins: data[1],
-//         distributedCoins: data[2],
-//         programs: {
-//           directProgram: data[3],
-//           bitcoinProgram: data[4],
-//           partnershipProgram: data[5],
-//           buildChallenge: data[6],
-//         },
-//       };
+function updateApiLumens() {
+  Promise.all([
+    commonLumens.totalCoins("https://horizon.stellar.org"),
+    commonLumens.availableCoins(),
+    commonLumens.distributionAll(),
+    commonLumens.distributionDirectSignup(),
+    commonLumens.distributionBitcoinProgram(),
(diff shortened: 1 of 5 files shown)


backend/lumens.js
```diff
@@ -1,54 +1,51 @@
...
+    commonLumens.distributionBuildChallenge(),
+    commonLumens.distributionUserAcquisition(),
+  ])
+    .then(function([
+      totalCoins,
+      availableCoins,
+      directDevelopment,
+      ecosystemSupport,
+      useCaseInvestment,
+      userAcquisition,
```
piyalbasu: these don't align with the Promise.all above, but i presume the callback params should look something like this

common/lumens.js
```diff
@@ -11,14 +11,31 @@ const horizonLiveURL = "https://horizon.stellar.org";
...
+    newProducts: "GCPWKVQNLDPD4RNP5CAXME4BEDTKSSYRR4MMEL4KG65NEGCOGNJW7QI2",
+  },
+  userAcquisition: {
+    inAppDistribution:
+      "GDKIJJIKXLOM2NRMPNQZUUYK24ZPVFC6426GZAEP3KUK6KEJLACCWNMX",
+    inAppDistributionHot:
+      "GAX3BRBNB5WTJ2GNEFFH7A4CZKT2FORYABDDBZR5FIIT3P7FLS2EFOZZ",
+    marketingSupport:
+      "GBEVKAYIPWC5AQT6D4N7FC3XGKRRBMPCAMTO3QZWMHHACLHTMAHAM2TP",
+  },
```
piyalbasu: my thought is to organize these uuid's in this fashion so it clearly matches the spec Kyle laid out and is clear for future devs

piyalbasu: this does, however, break `sdfAccounts`, but my guess is Jee will be making a change to that method to not simply iterate over account

common/lumens.js
```diff
@@ -86,6 +103,33 @@ export function distributionPartnershipProgram() {
   );
 }
 
+function sumRelevantAccounts(accountsObj) {
```
piyalbasu: created this little helper method to iterate over and sum an obj of accounts as all the programs I was dealing with had to do that

frontend/components/DistributionProgress.js
```diff
@@ -43,24 +43,32 @@ export default class DistributionProgress extends React.Component {
...
         directSignup,
         bitcoinProgram,
         partnershipProgram,
         buildChallenge,
+        ecosystemSupport,
+        useCaseInvestment,
+        userAcquisition,
       ] = results;
+      const sdfAccounts = 1000000;
+      const availableCoins = 1000000;
```
piyalbasu: just stubbing out some values atm to keep from breaking

frontend/components/DistributionProgress.js
```diff
@@ -99,39 +107,21 @@ export default class DistributionProgress extends React.Component {
...
-        name: "Direct signup program left",
-        value: parseInt(directSignupLeft),
-        color: toDistributeColor,
-      });
-      dataOuter.push({
-        name: "Partnership program left",
-        value: parseInt(partnershipProgramLeft),
-        color: toDistributeColor,
-      });
-
```
piyalbasu: moving the 3 programs I was working with into the pie chart. do we need to update the table underneath?

backend/lumens.js
```diff
@@ -1,54 +1,51 @@
 import * as commonLumens from "../common/lumens.js";
 
-let cachedData = {
```
piyalbasu: reverting all of Bartek's changes

ire-and-curses:
@piyalbasu  FYI for future PRs you can raise them as a [draft](https://github.blog/2019-02-14-introducing-draft-pull-requests/), which telegraphs this is WIP and makes it impossible to merge. Unfortunately you can't make a full PR into a draft retroactively.

common/lumens.js
jeesunikim: I was thinking of keeping Kyle's spec only in the API Changes (`backend/lumens.js`) and keep this object flat, like Bartek did previously so that if we need to group them, we do that in a function call later (for example, `distributionAll` function) and call that in `backend/lumens.js` so that we won't need to create an additional helper function to sum all the of balances since the final API only got the sum balances.

piyalbasu: ahh yes, that makes sense

jeesunikim:
Closing this PR since this has been merged (separately in my branch)

PR #129, opened 2019-11-15
update/remove_nodes_be > remove <Nodes/> related backend and frontend

jeesunikim:
* Removing <Nodes/> related backend and frontend files
-- I believe docs' node page was relying on `common/nodes.js`. We removed [nodes page](https://www.stellar.org/developers/guides/nodes.html) from docs; therefore `common/nodes.js` can be removed **unless there are other projects that rely on this**
* Updated `Readme`


bartekn:
LGTM but can we first update http://github.com/stellar/developers to remove "Known Validators" section. I'm afraid the build will fail if we merge this PR first.

vcarl:
I believe that was done already @bartekn in https://github.com/stellar/developers/pull/129, is there more?

bartekn:
@vcarl you're right, I missed that one!

PR #136, opened 2020-02-24
commented out directDevelopmentHot3

jeesunikim:
Removed directDevelopmentHot3 since it got merged into directDevelopmentHot2


kylemccollom:
yes. it's a great quick fix. we'll do a pass at a later date to tidy things up.

PR #137, opened 2020-03-10
Use horizon.stellar.org instead of horizon-mon.

abuiles:
We''ll be deploying Horizon 1.0.0 to horizon-mon soon and it could incorrectly display the network as down while it reingests state.

Let's use horizon.stellar.org for now and then we'll switch it back once we have successfully deployed 1.0.0 to mon.


abuiles:
@brahman81 it seems I'm not authorized to merge this PR, could you do it?

PR #138, opened 2020-03-11
Point horizon to mon, fix fee stats and changed operations.

abuiles:


frontend/components/FeeStats.js
```diff
@@ -102,9 +103,18 @@ export default class FeeStats extends React.Component {
             ) : (
               this.nameMap.map((field) => {
                 let styleFn = this.feeStyle;
-                let val = this.state.stats[field.id];
-                let displayVal = this.state.stats[field.id];
-                if (field.id == "ledger_capacity_usage") {
+                let stats = this.state.stats;
+                let id = field.id;
+
+                if (field.id.indexOf("max_fee.") === 0) {
```
vcarl: Another option here might be to use e.g. `_.get`, so the `id` can be interpreted as any key? Not a big deal though.

abuiles: oh cool, I didn't notice we have lodash already, let's do that :)

PR #139, opened 2020-03-18
Use horizon time to overcome inaccurate local time

### LINKED ISSUE #88: use horizon time
The delta between current time and last ledger close time uses the browser's clock. That's not always accurate. We should use a more accurate timestamp, perhaps from horizon.

tomerweller:
closes #88

### DIFF
--- a/frontend/common/time.js
+++ b/frontend/common/time.js
@@ -12,7 +12,7 @@ export function agoSeconds(ts) {
   return moment().diff(ts, "seconds") - _offset;
 }
 
-// calculate and format the time delta between now and ts
+// calculate and format the human-readable time delta between now and ts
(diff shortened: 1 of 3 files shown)


frontend/common/time.js
```diff
@@ -1,7 +1,20 @@
...
+export function setTimeOffset(offset) {
+  _offset = offset;
+}
+
+// calculate the time delta between now and ts in seconds
+export function agoSeconds(ts) {
+  return moment().diff(ts, "seconds") - _offset;
+}
+
+// calculate and format the time delta between now and ts
```
vcarl: Tiny nit: maybe
```suggestion
// calculate and format the human-readable time delta between now and ts
```

PR #149, opened 2020-08-25
Fix to displays the correct fee stats in the dashboard

marcinx:
The dashboard currently uses `max_fee` from https://horizon.stellar.org/fee_stats to calculate fee stats. It should be using `fee_charged` instead to reflect the actual fees that were charged in the ledger (instead of the maximum fees acceptable by submitting senders).

@abuiles could you check?


accordeiro:
@marcinx the changes look good, but there's something weird with the commit history (4 commits, some of them from 2018). Could you update this branch to only contain a commit with the proposed changes? Thanks :)

PR #150, opened 2020-08-26
Fix to display the correct fee stats in the dashboard

marcinx:
@accordeiro Cheers!


bartekn:
I'd vote for showing `fee_charged` in a separate panel or showing both values. The advantage of `max_fee` is that it shows you the bids so it's helpful for setting a fee for your next transaction.

accordeiro:
That makes sense, but it's a bigger change – @bartekn do you feel like replacing what we currently have with `fee_charged` will be misleading?

bartekn:
I don't have any data for users behaviour so I'm not sure if this is used by users to set fees. However, `fee_charged` gives you very little information about the fee auction.

However, what I agree is misleading are rows' names: `30th Percentile Accepted Fee` - is actually not Accepted Fee but fee bid.

Either way, changing the values without changing the panel name or rows' names can be misleading.

shanzzam:
@bartekn and @marcinx We're discussing the proposed changes and figuring out a way to make this less misleading. You may not see any movement on this for another week.

shanzzam:
UPDATED! Sooner rather than mentioned yesterday. @bartekn @marcinx @accordeiro 
Looking at feedback, what are the thoughts on these changes? Adding in an extra column and displaying both fee_charged and max_charged, as well as removing the "accepted" wording. 

Fee Stats Panel

accordeiro:
I think that's a great solution! @marcinx how do you feel about addressing that change? I don't want to put more on your plate, if you'd prefer we'd be happy to take it from here :)

PR #158, opened 2021-01-29
Make XLM balance calc resilient to deleted accounts

accordeiro:
## What?

Whenever computing the XLM supplies, if an account has been deleted, we should still be able to compute the overall number without errors.

## Why? 
When checking for the info of an account that has been deleted via Horizon, you get a 404 response. This is interpreted by axios as an error, and we should be able to handle that error gracefully (as a deleted account should simply be considered as having a zero XLM balance).

### DIFF
--- a/common/lumens.js
+++ b/common/lumens.js
@@ -61,7 +61,7 @@ export function getLumenBalance(horizonURL, accountId) {
     })
     .catch((error) => {
       if (error.response && error.response.status == 404) {
-        return "0.0"; // disconsider the balance of an account if there's an error with the request
+        return "0.0"; // consider the balance of an account zero if the account does not exist or has been deleted from the network
       } else throw error; // something else happened, and at this point we shouldn't trust the computed balance
     });
 }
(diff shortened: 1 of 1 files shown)


accordeiro:
Note: I only changed the `catch` part, all other cosmetic changes were automatically applied by the code formatter via husky webhook :)

accordeiro:
(I don't think the PR preview is working, seems to be pointing to master)

common/lumens.js
```diff
@@ -50,13 +50,20 @@ const accounts = {
...
+    .then((response) => {
+      var xlmBalance = find(
+        response.data.balances,
+        (b) => b.asset_type == "native",
+      );
+      return xlmBalance.balance;
+    })
+    .catch((error) => {
+      if (error.response && error.response.status == 404) {
+        return "0.0"; // disconsider the balance of an account if there's an error with the request
```
leighmcculloch, line 64: 💡 Suggested rewording of this comment. I don't think disconsider is a word, and I think we can be specific about the cases this occurs in.
```suggestion
        return "0.0"; // consider the balance of an account zero if the account does not exist or has been deleted from the network
```

leighmcculloch (dismissed):
One suggestion (💡), but regardless :shipit:.

common/lumens.js
accordeiro: 💯 good catch :) I meant disregard, but I like your phrasing better.

acharb (approved):
LGTM!

leighmcculloch (approved):

PR #164, opened 2021-04-26
Switch build image to ubuntu 20.04

jacekn:
* switch to ubuntu 20.04 base image
* move away from deprecated "apt-key add"
* use --no-install-recommends to reduce number of packages installed
* add --pull to the Makefile to ensure we always use latest images


brahman81 (approved):
LGTM

Issue #181, opened 2021-11-09
Feature Request: Improvements to the lumens API


satyamz:
### What problem does your feature solve?
When we use `/api/v2/lumens` API it returns values in the float format. It would be very useful to have values of `/api/v2/lumens` API to be reported as a whole numbers, to achieve that we can multiply current number with `10^7`. We can consider adding a new filter to the API to return values in the different format.

Currently, programming languages such as javascript (old version), python etc. truncate the fraction part of the number and that makes it very difficult to get exact `totalSupply` number. It'd be great to have `totalSupply` and other fields provided as whole numbers.

### What would you like to see?
In addition to the whole number support, it'd also be useful to have `totalSupply` sanity check inbuilt in the API. API should return following two numbers:
1. `lumensSupply` = `totalSupply` - `feePool`
2.  `circulatingLumensSupply`= `upgradeReserve` + `circulatingSupply` + `sdfMandate`

Above two numbers should be same. This will ensure that `totalSupply` will always be equal to sum of four categories `upgradeReserve`, `circulatingSupply`, `sdfMandate` and `feePool`.

(Nomenclature for the above two numbers can be different too.)

### What alternatives are there?
None!

acharb:
@satyamz with this merged for the Blazemeter Total Lumens test the endpoints to compare are:
`/api/v3/lumens/total-supply-check` and
`/api/v3/lumens/total-supply`

and can be compared with the strings directly so no parsing is needed

PR #182, opened 2021-11-12
{backend, common}: open new total-supply-check endpoint

acharb:
**WHAT**
adds a new endpoint `/api/v3/lumens/total-supply-check`. This endpoint returns the sum of:
`upgrade_reserve + fee_pool + sdf_mandate + circulating_supply`

also adds a new endpoint `/api/v3/lumens/total-supply`. The v2 version of this endpoint returned the value as a Number, which would lose a decimal point of precision. So the v3 endpoints return as strings

**WHY**
right now Blazemeter is using dashboard endpoints for the [Total Lumens check](https://www.runscope.com/radar/v60ug3n4szv3/c5acd368-d46c-4a9f-a5f5-4f08f96fbe6e). In that check we're using the JS `parseFloat` and `parseInt` methods, which lose precision when the number is larger than what js `Number` can handle.

By returning this new endpoint, Blazemeter only has to compare `/api/v3/lumens/total-supply` to `/api/v3/lumens/total-supply-check`, so it doesn't have to do any parsing/multiplication on its end. And all values are handled by `BigNumber` on the dashboard side. Which is good because Blazemeter has limited [library support](https://guide.blazemeter.com/hc/en-us/articles/360001858778).

closes https://github.com/stellar/dashboard/issues/181

note: the original request was for returning whole numbers. I think this solution is better for the reason stated above that Blazemeter would not be able to handle the whole number anyways since `parseInt` would lose precision. So comparing the strings directly instead should be more accurate.

### DIFF
--- a/backend/app.js
+++ b/backend/app.js
@@ -33,15 +33,17 @@ if (process.env.DEV) {
 
 app.get("/api/ledgers/public", ledgers.handler);
 app.get("/api/lumens", lumens.handler);
-app.get("/api/v2/lumens", lumensV2.handler);
 
+app.get("/api/v2/lumens", lumensV2.handler);
 /* For CoinMarketCap */
 app.get("/api/v2/lumens/total-supply", lumensV2.totalSupplyHandler);
 app.get("/api/v2/lumens/circulating-supply", lumensV2.circulatingSupplyHandler);
 
+app.get("/api/v3/lumens", lumensV3.handler);
 app.get("/api/v3/lumens/total-supply-check", lumensV3.totalSupplySumHandler);
-app.get("/api/v3/lumens/circulating-supply", lumensV3.circulatingSupplyHandler);
+/* For CoinMarketCap */
 app.get("/api/v3/lumens/total-supply", lumensV3.totalSupplyHandler);
+app.get("/api/v3/lumens/circulating-supply", lumensV3.circulatingSupplyHandler);
 
 app.listen(app.get("port"), function() {
   console.log("Listening on port", app.get("port"));
--- a/backend/app.js
+++ b/backend/app.js
@@ -4,6 +4,7 @@ import logger from "morgan";
 
 import * as lumens from "./lumens.js";
 import * as lumensV2 from "./v2/lumens.js";
+import * as lumensV3 from "./v3/lumens.js";
 import * as ledgers from "./ledgers.js";
 
 var app = express();
@@ -38,13 +39,9 @@ app.get("/api/v2/lumens", lumensV2.handler);
 app.get("/api/v2/lumens/total-supply", lumensV2.totalSupplyHandler);
 app.get("/api/v2/lumens/circulating-supply", lumensV2.circulatingSupplyHandler);
 
-app.get("/api/v2/lumens/total-supply-check", lumensV2.totalSupplySumHandler);
+app.get("/api/v3/lumens/total-supply-check", lumensV3.totalSupplySumHandler);
 
-// Temporary endpoint, returning string instead of number for precision
-app.get(
-  "/api/v2/lumens/total-supply-full",
-  lumensV2.totalSupplyFullDataHandler,
-);
+app.get("/api/v3/lumens/total-supply", lumensV3.totalSupplyDataHandler);
 
 app.listen(app.get("port"), function() {
   console.log("Listening on port", app.get("port"));
--- a/backend/app.js
+++ b/backend/app.js
@@ -41,7 +41,7 @@ app.get("/api/v2/lumens/circulating-supply", lumensV2.circulatingSupplyHandler);
 
 app.get("/api/v3/lumens/total-supply-check", lumensV3.totalSupplySumHandler);
 app.get("/api/v3/lumens/circulating-supply", lumensV3.circulatingSupplyHandler);
-app.get("/api/v3/lumens/total-supply", lumensV3.totalSupplyDataHandler);
+app.get("/api/v3/lumens/total-supply", lumensV3.totalSupplyHandler);
 
 app.listen(app.get("port"), function() {
   console.log("Listening on port", app.get("port"));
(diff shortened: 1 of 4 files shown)


common/lumens.js
```diff
@@ -220,3 +220,18 @@ export function circulatingSupply() {
...
 }
+
+export function totalSupplySum() {
+  return Promise.all([
+    getUpgradeReserve(),
+    feePool(),
+    sdfAccounts(),
+    circulatingSupply(),
+  ]).then((balances) => {
+    return reduce(
```
accordeiro, line 231: Fancy! :)

common/lumens.js
```diff
@@ -220,3 +220,18 @@ export function circulatingSupply() {
...
+      balances,
+      (sum, balance) => sum.add(balance),
```
accordeiro, line 233: ❓ - do we also need to convert the computation of each of the promise functions (getUpgradeReserve, feePool, etc) to BigNumber so we don't lose precision? Probably not, but thought worth checking.

accordeiro (approved):
Looks good to me! Left a small question inline

backend/v2/lumens.js
```diff
@@ -60,6 +68,8 @@ function updateApiLumens() {
       totalSupplyData = response.totalSupply * 1;
       circulatingSupplyData = response.circulatingSupply * 1;
 
+      totalSupplySumData = totalSupplySum * 1;
```
leighmcculloch, line 71: ❔ What's the significance of the `* 1`?

leighmcculloch (dismissed):
One non-blocking question (❔). Looks great!

common/lumens.js
acharb: good point, since not all the values from the promise are returned as bignumber I changed just to be safe
https://github.com/stellar/dashboard/pull/182/commits/06e4a72d327ccb7091b82a464212cc4a9a18ffce

backend/v2/lumens.js
piyalbasu: Is this possibly type coercion? I see this in a couple places in this repo

acharb: I missed that that's actually converting the values to number when storing it. It actually loses a decimal point. Changed to keep as string here:
https://github.com/stellar/dashboard/pull/182/commits/a9d56abdb4f8558e7747f5be6bb23160c2f1bae3

@accordeiro these two endpoints are now returned as string instead of number to save precision. Looks like they're for CoinMarketCap. Is this something we should notify them about?
```
/* For CoinMarketCap */
app.get("/api/v2/lumens/total-supply", lumensV2.totalSupplyHandler);
app.get("/api/v2/lumens/circulating-supply", lumensV2.circulatingSupplyHandler);
```

accordeiro: Yes, I think we should notify. But to prevent any breakage in the meantime, instead of replacing fields, can we just add **new** fields with correct precision values?

accordeiro: Or a `/v3` API namespace

leighmcculloch: A `/v3` API namespace would allow us to track usage.

acharb: sgtm
https://github.com/stellar/dashboard/pull/182/commits/d072a589c5f1951acf7e842c3a2cd7f778711738

backend/app.js
```diff
@@ -38,6 +39,9 @@ app.get("/api/v2/lumens", lumensV2.handler);
 app.get("/api/v2/lumens/total-supply", lumensV2.totalSupplyHandler);
 app.get("/api/v2/lumens/circulating-supply", lumensV2.circulatingSupplyHandler);
```
leighmcculloch, line 40: ❓ Why don't we have the circulating supply and `/api/v2/lumens` on the `v3`? If we keep functionality only available in old versions we may struggle to get people off old versions.

backend/v3/lumens.js
```diff
@@ -0,0 +1,30 @@
+import * as commonLumens from "../../common/lumens.js";
+import BigNumber from "bignumber.js";
+
+let totalSupplyData;
+let totalSupplySumData;
+
+export const totalSupplyDataHandler = function(req, res) {
```
leighmcculloch, line 7: 💡 Could we name this totalSupplyHandler? I don't think the `Data` helps communicate the intent of what this does anymore than it did without it, and it can be helpful in API upgrades if we preserve as much as possible of the old terminology for things that do not matter or do not need to change.

leighmcculloch (commented):
One question (❓) and one non-blocking suggestion (💡).

backend/app.js
acharb: agreed this was needed to https://github.com/stellar/dashboard/pull/182/commits/85d77b3767787db56bf755c998ddf44cb630f619

leighmcculloch: What about `/api/v2/lumens`?

backend/v3/lumens.js
acharb: https://github.com/stellar/dashboard/pull/182/commits/87afef9d9eb0fb3d97d698dcab26ff2660f2cb7a

backend/app.js
leighmcculloch: ☝🏻

acharb: oh hmm it would be exactly the same. if we make that one v3 does it imply something changed?

leighmcculloch: We should have every endpoint in the new api version, otherwise we are saying anything not migrated is deprecated.

acharb: I'm not sure if these are all or nothing changes. Eg. just updating the endpoints that changed makes sense in my head. But I'll defer if you think that one should be upped a version too

leighmcculloch: I assume `/api/ledgers/public` was deprecated when we released v2. But if it wasn't, we should include it too. @accordeiro Do you know?

acharb: > I assume /api/ledgers/public was deprecated when we released v2. But if it wasn't, we should include it too. @accordeiro Do you know?

note: I'll be working on updating the backend of this repo so I could do this cleanup/add to v3 as part of that

leighmcculloch (approved):

PR #183, opened 2021-11-15
backend: use single endpoint for checking total supply

acharb:
**WHAT**
uses a single endpoint to check the total supply values instead of 2

**WHY**
the Runscope alert is still throwing errors. The cause may be due to Runscope calling 2 endpoints separately causing a race condition.
(more context here: https://stellarfoundation.slack.com/archives/CANJGG94K/p1636975310074200)

### DIFF
--- a/backend/app.js
+++ b/backend/app.js
@@ -40,7 +40,7 @@ app.get("/api/v2/lumens/total-supply", lumensV2.totalSupplyHandler);
 app.get("/api/v2/lumens/circulating-supply", lumensV2.circulatingSupplyHandler);
 
 app.get("/api/v3/lumens", lumensV3.handler);
-app.get("/api/v3/lumens/total-supply-check", lumensV3.totalSupplyCheckHandler);
+app.get("/api/v3/lumens/all", lumensV3.totalSupplyCheckHandler);
 /* For CoinMarketCap */
 app.get("/api/v3/lumens/total-supply", lumensV3.totalSupplyHandler);
 app.get("/api/v3/lumens/circulating-supply", lumensV3.circulatingSupplyHandler);
--- a/backend/app.js
+++ b/backend/app.js
@@ -40,7 +40,7 @@ app.get("/api/v2/lumens/total-supply", lumensV2.totalSupplyHandler);
 app.get("/api/v2/lumens/circulating-supply", lumensV2.circulatingSupplyHandler);
 
 app.get("/api/v3/lumens", lumensV3.handler);
-app.get("/api/v3/lumens/total-supply-check", lumensV3.totalSupplySumHandler);
+app.get("/api/v3/lumens/total-supply-check", lumensV3.totalSupplyCheckHandler);
 /* For CoinMarketCap */
 app.get("/api/v3/lumens/total-supply", lumensV3.totalSupplyHandler);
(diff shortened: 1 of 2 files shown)


backend/v3/lumens.js
```diff
@@ -77,5 +87,5 @@ function updateApiLumens() {
     });
 }
 
-setInterval(updateApiLumens, 10 * 60 * 1000);
+setInterval(updateApiLumens, 1 * 60 * 1000);
```
acharb, line 90: changed this to update every minute for faster debugging. Once confirmed this fix works I'll change back to every 10 minutes

backend/app.js
```diff
@@ -40,7 +40,7 @@ app.get("/api/v2/lumens/total-supply", lumensV2.totalSupplyHandler);
 app.get("/api/v2/lumens/circulating-supply", lumensV2.circulatingSupplyHandler);
 
 app.get("/api/v3/lumens", lumensV3.handler);
-app.get("/api/v3/lumens/total-supply-check", lumensV3.totalSupplySumHandler);
+app.get("/api/v3/lumens/total-supply-check", lumensV3.totalSupplyCheckHandler);
```
leighmcculloch, line 43: ❗ Are we open to changing the name of this endpoint? `total-supply-check` doesn't really describe the data, since it is more than total supply, and `check` is a verb which makes it sound like calling this endpoint does something.

I suggest `/api/v3/lumens/all`.

backend/app.js
```diff
@@ -40,7 +40,7 @@ app.get("/api/v2/lumens/total-supply", lumensV2.totalSupplyHandler);
...
 /* For CoinMarketCap */
 app.get("/api/v3/lumens/total-supply", lumensV3.totalSupplyHandler);
 app.get("/api/v3/lumens/circulating-supply", lumensV3.circulatingSupplyHandler);
```
leighmcculloch, line 46: 💡 I think we can remove these two endpoints and as part of `v3` deprecate the old v2 endpoints and tell people to use the new endpoint. The new endpoint above contains all the same data and if we're introducing a new version of the API we can deprecate duplicate data points. Thoughts @acharb @accordeiro?

leighmcculloch (commented):
One request (❗), and a suggestion (💡).|

backend/app.js
acharb: the purpose of this endpoint is really just for runscope. I only added all the fields so that if it fails runscope will get all the numbers that TotalSupply is calculated from. `all` to me sounds like it's a better version of `/lumens` (atm I guess it is) that people might use instead. I can change to `all` for now, I can't think of a better name for its current use

leighmcculloch: Do we have API documentation for the dashboard?

backend/app.js
acharb: makes sense to me. I'm not sure why we had the 2 single endpoints for CoinMarketCap in v2 though, since the `api/v2/lumens` also had that data too. I'll create a ticket to track that. In the mean time for this PR can we keep to debug the runscope error?

backend/app.js
acharb: ya  possibly we can combine into a single endpoint. But I think we should make sure this single endpoint check fixes the Runscope error. If not it will have to change anyways

leighmcculloch (approved):

PR #185, opened 2021-11-16
{backend, common}: compute the derived values from updateApiLumens

acharb:
**WHAT**
change the derived values (totalSupply, circulatingSupply, totalSupplySum) to be computed in the updateApiLumens method, instead of the promises returned from common.

**WHY**
the derived values each call Horizon on their own (eg. `totalSupplySum` calls `feePool` and `circulatingSupply` calls `feePool` through `nonCirculatingSupply`). The promises calling Horizon on their own create a race condition, which is causing an error when testing that the values are the same with Runscope. So instead, calculate the values in the final method instead, so each Horizon endpoint is called just once.

<hr>

Also:
In this PR you can see that `totalSupply` will always equal `totalSupplySum` (if you plug in circulatingSupply to the totalSupplySum equation). So currently Dashboard is just checking for the race condition that I'm removing now, since how Dashboard computes `circulatingSupply` is derived from `totalSupply`. So, for checking that `totalSupply` is correct, I think we should use Hubble. That should be a pretty simple query in hubble to get the `circulatingSupply`.

@satyamz said that Runscope previously was using the `core-mon-api` [`/accounts' endpoint](https://github.com/stellar/core-mon-api/blob/983a9765f36c0d3e47354232549ffac4a1465b52/index.js#L22-L34), which was querying all of the  accounts w/ lumens and signers. Dashboard can't make this kind of request.

Alternatively, dashboard can call Horizon's [totalCoins](https://horizon.stellar.org/ledgers/?order=desc&limit=1) endpoint and subtract the `inflationAccount` lumens, to make sure it's always equal to 1B xlm.

Lmk what you think @accordeiro @leighmcculloch

### DIFF
--- a/backend/v3/lumens.js
+++ b/backend/v3/lumens.js
@@ -90,5 +90,5 @@ function updateApiLumens() {
     });
 }
 
-setInterval(updateApiLumens, 8 * 1000);
+setInterval(updateApiLumens, 1 * 60 * 1000);
 updateApiLumens();
--- a/backend/v3/lumens.js
+++ b/backend/v3/lumens.js
@@ -40,8 +40,7 @@ function updateApiLumens() {
       upgradeReserve,
       sdfMandate,
     ]) {
-      let totalSupply = new BigNumber(0)
-        .plus(originalSupply)
+      let totalSupply = new BigNumber(originalSupply)
         .plus(inflationLumens)
         .minus(burnedLumens);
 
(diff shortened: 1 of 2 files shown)


leighmcculloch (approved):

acharb:
going to merge as this PR fixes the race condition. As for improving the total lumen supply check hopefully this can be done with a new horizon endpoint: https://github.com/stellar/go/issues/4083

PR #188, opened 2021-12-01
{backend, test}: add initial backend tests

acharb:
**WHAT**
Add testing to the dashboard backend.

**WHY**
Testing will help prevent future bugs while this app is getting refactored. Adding the packages and a simple test at first but more test coverage will probably be added/needed in the future.

Also moved the interval calls of `updateApiLumens`, which updates the cache, to be called in app.js. This is so when importing the functions to be tested the functions are not invoked, as it is currently doing.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,6 +41,7 @@
     "bourbon": "^4.3.4",
     "bourbon-neat": "^2.0.0",
     "browser-sync": "^2.18.2",
+    "chai": "^4.3.4",
     "css-loader": "2.1.1",
     "dompurify": "^2.2.6",
     "express": "^4.14.0",
--- a/package.json
+++ b/package.json
@@ -41,7 +41,6 @@
     "bourbon": "^4.3.4",
     "bourbon-neat": "^2.0.0",
     "browser-sync": "^2.18.2",
-    "chai": "^4.3.4",
     "css-loader": "2.1.1",
     "dompurify": "^2.2.6",
     "express": "^4.14.0",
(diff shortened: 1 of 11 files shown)


acharb:
Not sure what these Netlify checks are, but they aren't blocking merge, and Ops is looking into them
https://stellarfoundation.slack.com/archives/C02U19A2A/p1638465561444900 cc @tprobinson

package.json
```diff
@@ -51,6 +53,7 @@
     "html-react-parser": "^1.2.4",
     "json-loader": "^0.5.1",
     "lodash": "^4.17.15",
+    "mocha": "^8.4.0",
```
leighmcculloch, line 56: ❓ I'm not very familiar with how dependencies work, so I could be way off the mark with this, but since mocha is a test framework, should it be in the `devDependencies`?

test/unit/backend.js
```diff
@@ -0,0 +1,31 @@
...
+  });
+});
+
+describe("lumens v3", function() {
+  describe("updateApiLumens", function() {
+    it("should run without error", async function() {
+      let err = await v3.updateApiLumens();
+      chai.assert.isUndefined(err, "there was no error");
+    });
+  });
```
leighmcculloch, line 30: ❓ Are there other things we can test these functions on, like some side-effect they have?

leighmcculloch (approved):
I left two questions, but neither blocking and it looks good to me.

test/unit/backend.js
acharb: yup! I plan on adding tests for those. Didn't want to make this PR too big

package.json
acharb: ya  I wasn't sure if the github action environment would install devDependencies, but looks like it does 👍  moved
https://github.com/stellar/dashboard/pull/188/commits/cc7de972a00f5134ffdd4bc92cca6ceffcfb27c4

acharb:
@leighmcculloch can you approve again? I'll change the dismiss stale review flag in terraform

leighmcculloch (approved):
🎉

PR #191, opened 2021-12-03
backend: make cache update methods wait

acharb:
**WHAT**
make the cache update calls synchronous

**WHY**
there's 3 versions of the `updateApiLumens` method being called simultaneously, and each hits Horizon multiple times. This is causing a 429 error with Horizon. So make each one wait for the other. The calls are quick so they don't wait long.

note: looking at kibana each version of the api is being called (last I looked at least ~1000 a month each), so can't deprecate versions atm

### DIFF
--- a/backend/app.js
+++ b/backend/app.js
@@ -49,10 +49,10 @@ app.listen(app.get("port"), function() {
   console.log("Listening on port", app.get("port"));
 });
 
-export function updateLumensCache() {
-  lumens.updateApiLumens();
-  lumensV2.updateApiLumens();
-  lumensV3.updateApiLumens();
+export async function updateLumensCache() {
+  await lumens.updateApiLumens();
+  await lumensV2.updateApiLumens();
+  await lumensV3.updateApiLumens();
 }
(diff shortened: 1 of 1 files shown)


leighmcculloch (approved):

backend/app.js
```diff
@@ -49,10 +49,10 @@ app.listen(app.get("port"), function() {
...
 });
 
-export function updateLumensCache() {
-  lumens.updateApiLumens();
-  lumensV2.updateApiLumens();
-  lumensV3.updateApiLumens();
+export async function updateLumensCache() {
+  await lumens.updateApiLumens();
+  await lumensV2.updateApiLumens();
+  await lumensV3.updateApiLumens();
```
leighmcculloch, line 55: 💡 You could wait for them in parallel using `await Promise.all(...)`.

acharb: I actually need it to be not in parallel, so that all the calls don't hit horizon at the same time :/ but agreed otherwise that would be cleaner

PR #193, opened 2021-12-07
backend: adding integration tests

acharb:
**WHAT**
add integration tests for the backend api endpoints

**WHY**
to ensure they work as intended during code changes. Will help with project maintenance going forward.

note: the only endpoint not currently tested is `api/ledgers/public`, because it requires a postgres database. I followed [directions here](https://docs.github.com/en/actions/using-containerized-services/creating-postgresql-service-containers) but couldn't get it working correctly (though it is pretty cool), and the plan is to deprecate postgres for redis anyways so ignoring that endpoint until then

### DIFF
--- a/app.js
+++ b/app.js
@@ -2,4 +2,8 @@ require("babel-register")({
   presets: ["es2015"],
 });
 
-require("./backend/app.js");
+// Run backend with interval cache updates.
+const { updateLumensCache } = require("./backend/app.js");
+
+setInterval(updateLumensCache, 10 * 60 * 1000);
+updateLumensCache();
--- a/test/tests/integration/backend.js
+++ b/test/tests/integration/backend.js
@@ -2,22 +2,15 @@ import chai from "chai";
 const request = require("supertest");
 
 const { app, updateLumensCache } = require("../../../backend/app.js");
-const { updateResults, cachedData } = require("../../../backend/ledgers.js");
-
 describe("integration", function() {
   // update caches
   before(async function() {
     await updateLumensCache();
-    await updateResults();
   });
 
   describe("backend api endpoints", function() {
     it("should return data of correct type", async function() {
       let testCases = [
-        {
-          url: "/api/ledgers/public",
-          wantResultType: "array",
-        },
         {
           url: "/api/lumens",
           wantResultType: "object",
--- a/test/tests/integration/backend.js
+++ b/test/tests/integration/backend.js
@@ -9,62 +9,76 @@ describe("integration", function() {
   });
 
   describe("backend api endpoints", function() {
-    it("should return data of correct type", async function() {
-      let testCases = [
-        {
-          url: "/api/lumens",
-          wantResultType: "object",
-        },
-        {
-          url: "/api/v2/lumens",
-          wantResultType: "object",
-        },
-        {
-          url: "/api/v2/lumens/total-supply",
-          wantResultType: "number",
-        },
-        {
-          url: "/api/v2/lumens/circulating-supply",
-          wantResultType: "number",
-        },
-        { url: "/api/v3/lumens", wantResultType: "object" },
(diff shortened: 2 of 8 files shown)


app.js
```diff
@@ -2,4 +2,8 @@ require("babel-register")({
   presets: ["es2015"],
 });
 
-require("./backend/app.js");
+// Run backend with interval cache updates.
```
acharb, line 5: also moved this code to trigger the cache updates to the top level app.js, so it doesn't get called on integration tests

test/tests/integration/backend.js
```diff
@@ -0,0 +1,70 @@
...
+      ];
+
+      for (var tc of testCases) {
+        console.log("testing: ", tc.url);
+        let { body } = await request(app)
+          .get(tc.url)
+          .expect(200);
+
+        chai.expect(body).to.be.an(tc.wantResultType);
+        chai.expect(containsData(body)).to.be.false;
```
leighmcculloch, line 49: ❔ Why do we want all the endpoints to contain no data?

test/tests/integration/backend.js
```diff
@@ -0,0 +1,70 @@
...
+
+const { app, updateLumensCache } = require("../../../backend/app.js");
+describe("integration", function() {
+  // update caches
+  before(async function() {
+    await updateLumensCache();
+  });
+
+  describe("backend api endpoints", function() {
+    it("should return data of correct type", async function() {
```
leighmcculloch, line 12: 💡 I think these tests will be easier to maintain if there's a single test per endpoint. This test is a sort of mini-framework for a specific type of test, but it will probably be approximately the same effort to write them out one by one, and easier to modify then in each case. This is just my perspective though, and just a suggestion.

Example:
```js
   describe("backend api endpoints", function() {
     it("/api/lumens should return an empty object", async function() {
         let { body } = await request(app)
           .get("/api/lumens")
           .expect(200);
         chai.expect(body).to.be.an("object");
         chai.expect(Object.keys(data).length).to.eq(0);
     });
   });
 });
```

leighmcculloch (approved):
I left a suggestion and a question, but nothing blocking.

test/tests/integration/backend.js
acharb: whoops I switched it from `isEmpty` to `containsData` to get rid of the double negative but looks like forgot to switch the equalities as well. Fixed
https://github.com/stellar/dashboard/pull/193/commits/26a04a2ce3029ab350b462597d3ee48c6765089b

test/tests/integration/backend.js
acharb: agreed I like how they'll be easier to modify with your way 👍  changed
https://github.com/stellar/dashboard/pull/193/commits/7a9a3e5d4d21bc51729c179a8927a49fe43e0c30

test/tests/integration/backend.js
```diff
@@ -0,0 +1,84 @@
...
+  });
+
+  describe("backend api endpoints", function() {
+    it("/api/lumens should return successfuly with data", async function() {
+      let { body } = await request(app)
+        .get("/api/lumens")
+        .expect(200);
+
+      chai.expect(body).to.be.an("object");
+      chai.expect(Object.keys(body).length).to.not.equal(0);
```
leighmcculloch, line 18: Is there a consistent set of keys we expect in this object?

acharb: yes, I'm checking them in the unit test
https://github.com/stellar/dashboard/blob/eecbe6641f9ee10dc7d26b80d0398839fe6d3e00/test/tests/unit/backend.js#L5-L31
I didn't think it needed to be checked again in the integration test

leighmcculloch: Ah got it. 👍🏻

Issue #196, opened 2021-12-07
Update Node Version to LTS 16


acharb:
Currently is version 10, which is end of life. Will involve updating dependencies as well

PR #199, opened 2021-12-08
all: update node engine to v16

acharb:
**WHAT**
upgrading the Nodejs engine to LTS, v16

**WHY**
the engine was currently 10, which hasn't had active support for about a [year and a half](https://endoflife.date/nodejs). Upgrading to v16 will help with the current repo, and also allow for new packages that get added as this [repo gets refactored](https://stellarorg.atlassian.net/wiki/spaces/PET/pages/4244078593/Dashboard+Backend+Upgrade+Plan) to be at a good state from the beginning.

closes https://github.com/stellar/dashboard/issues/196

### Some notes about the dependencies that needed updating:
- gulp ([gulp 3 is broken](https://github.com/gulpjs/gulp/issues/2324) for Node versions above 10)
  - this involved some refactoring of how gulp uses the tasks
- node-sass 6.0.1 is [minimum for node16](https://stackoverflow.com/a/67303155/8846296)
- pg - needs [8.2 or above](https://node-postgres.com/) for node16
  - this needed to add a `rejectUnauthorized: false` flag ([source](https://stackoverflow.com/questions/61350186/how-to-solve-the-database-connection-error-sequelizeconnectionerror)) to the config like below otherwise a `self-signed certificate` error is thrown

```
ssl: {
            require: true,
            rejectUnauthorized: false,
          },
```
I wasn't sure how safe this change is, but from my research online this is the most often used and best solution to pg v8+ connecting to the db.  It is also temporary as postgres will be removed soon in favor of redis.


backend/postgres.js
```diff
@@ -4,7 +4,17 @@ export const sequelize = new Sequelize(
...
     : process.env.POSTGRES_URL,
-  process.env.DEV ? {} : { dialect: "postgres", dialectOptions: { ssl: true } },
+  process.env.DEV
+    ? {}
+    : {
+        dialect: "postgres",
+        dialectOptions: {
+          ssl: {
+            require: true,
+            rejectUnauthorized: false,
```
leighmcculloch, line 14: ❓ What effect does adding `rejectedUnauthorized: false` have? Does it downgrade any security we have today? I couldn't figure out what it means.

leighmcculloch: Answering my own question, it appears `rejectedUnauthorized: false` means an `sslmode` requiring verification of the server certificate won't be used. This should be fine given that the default is `prefer` which requires encryption but not verification.

Refs:
- https://github.com/brianc/node-postgres/blob/4b229275cfe41ca17b7d69bd39f91ada0068a5d0/packages/pg-connection-string/README.md
- https://www.postgresql.org/docs/9.1/libpq-ssl.html

leighmcculloch (approved):
👏🏻

backend/postgres.js
leighmcculloch: Also, one more learning I had, is that if the connection string contains the sslmode, or any other ssl parameters, the ssl object being specified here will be ignored. So if in any production environment sslmode=verify-ca is being used, it will take precedence and this change should not be a downgrade.

Refs:
- >If you plan to use a combination of a database connection string from the environment and SSL settings in the config object directly, then you must avoid including any of sslcert, sslkey, sslrootcert, or sslmode in the connection string. If any of these options are used then the ssl object is replaced and any additional options provided there will be lost.

  https://node-postgres.com/features/ssl#usage-with-connectionstring

Issue #197, opened 2021-12-07
Add TypeScript to backend


acharb:
1. Add typescript to project
2. Refactor backend code to use

PR #202, opened 2021-12-17
all: add typescript to backend

acharb:
**WHAT**
add typescript support to the backend. For this PR just adding the initial typescript infra changes needed, following PRs will refactor the `backend/*.js` files to typescript.

part of https://github.com/stellar/dashboard/issues/197

**WHY**
using typescript with the typescript compiler will notify devs at compile time of type errors + reduce bugs. Also SDF js projects are primarily using typescript, so brings this project inline with our other ones.

note: for running the backend locally I chose to use `ts-node`, so that you can run it with one command and not have to create new files. But for production using the `tsc` compiler + `node` because `ts-node` is advised [not for production use](https://github.com/TypeStrong/ts-node/issues/104).


acharb:
@piyalbasu @quietbits (who I think may have an opinion on this)

I played around with [@babel/preset-typescript](https://babeljs.io/docs/en/babel-preset-typescript) to compile the backend `ts` using webpack (which I'm guessing the frontend may use when it adds typescript), but it wasn't trivial refactoring webpack to compile the backend too, so I decided to stay with `tsc` (which seems easier and we may want to use instead of babel on the frontend anyways). lmk if you have feelings against this approach

package.json
```diff
@@ -11,7 +11,8 @@
     "build": "gulp build",
     "heroku-postbuild": "gulp build",
     "lint:prettier": "prettier --config --write '**/*.js'",
-    "test": "DEV=true mocha './test/**/*.js' --exit"
+    "test": "DEV=true mocha './test/**/*.js' --exit",
+    "dev-backend": "UPDATE_DATA=true DEV=true ts-node backend/app.ts"
```
piyalbasu, line 15: really small - but in our JS repos, we generally namespace dev environment scripts as `start`. Possible to rename this script to `start:backend`?

piyalbasu (approved):

piyalbasu:
> @piyalbasu @quietbits (who I think may have an opinion on this)
> 
> I played around with [@babel/preset-typescript](https://babeljs.io/docs/en/babel-preset-typescript) (which I'm guessing the frontend may use when it adds typescript) to compile the backend using webpack , but it wasn't trivial refactoring webpack to compile the backend too, so I decided to stick with `tsc` (which seems much easier and we may want to use instead of babel on the frontend anyways). lmk if you have thoughts against this approach

This makes sense. Totally favor a simpler `tsc` script. Webpack stuff can get pretty hairy pretty fast. We'll prob eliminate Webpack altogether when we refactor the FE anyway

package.json
acharb: good point, `start:backend` sgtm 👍 , changed
https://github.com/stellar/dashboard/pull/202/commits/f2c1471c5b79981dfa6cb9a8558531e60d3727ab

PR #204, opened 2021-12-18
{backend, test}: change backend files and backend tests to typescript

acharb:
**WHAT**
converting backend files from JS to TS. Some slight refactoring was needed (pointed out in comments), but in general no functional changes.

closes https://github.com/stellar/dashboard/issues/197

**WHY**
Typescript will make the project more aligned with other SDF ones, as well as prevent bugs thanks to TS compiler


backend/postgres.js
```diff
@@ -1,71 +0,0 @@
...
-        dialectOptions: {
-          ssl: {
-            require: true,
-            rejectUnauthorized: false,
-          },
-        },
-      },
-);
-
-export const NodeMeasurement = sequelize.define(
```
acharb, line 20: removed this `NodeMeasurement` table, it's not being used

common/lumens.js
```diff
@@ -1,10 +1,10 @@
 // This file contains functions used both in backend and frontend code.
 // Will be helpful to build distribution stats API.
-import axios from "axios";
```
acharb, line 3: needed to remove babel on the backend because of a [bug with sequelize](https://github.com/sequelize/sequelize/issues/7840) (annoyingly since sequelize will be removed soon). So changed `common/lumens.js` to use commonJS `common/require` instead of `import/export` which works for both the frontend and backend

package.json
```diff
@@ -38,7 +38,7 @@
     "babel-preset-es2015": "^6.22.0",
     "babel-preset-react": "^6.23.0",
     "babel-register": "^6.24.0",
-    "bignumber.js": "^3.0.1",
+    "bignumber.js": "^9.0.1",
```
acharb, line 41: bignumber.js types added in version 4

common/lumens.js
```diff
@@ -1,10 +1,10 @@
...
-import BigNumber from "bignumber.js";
-import map from "lodash/map";
-import reduce from "lodash/reduce";
-import find from "lodash/find";
+const axios = require("axios");
+const BigNumber = require("bignumber.js");
+const map = require("lodash/map");
+const reduce = require("lodash/reduce");
+const find = require("lodash/find");
```
acharb, line 7: needed to remove babel on the backend because of a [bug with sequelize](https://github.com/sequelize/sequelize/issues/7840#issuecomment-315576088) (annoyingly since sequelize will be removed soon). So changed common/lumens.js to use commonJS common/require instead of import/export which works for both the frontend and backend

tsconfig.json
```diff
@@ -2,12 +2,14 @@
...
   "compilerOptions": {
     // temporary flags during js->ts transition
-    "allowJs": false, 
+    "allowJs": true, 
     "isolatedModules": false,
-
-    "target": "es5",      
-    "rootDir": "./backend",
-    "baseUrl": "./"
+    "target": "es6",
```
acharb, line 7: changing target to es6 because of [sequelize](https://github.com/sequelize/sequelize/issues/7840#issuecomment-317212324)

common/lumens.d.ts
```diff
@@ -0,0 +1,15 @@
+export function getLumenBalance(horizonURL: any, accountId: any): Promise<any>;
```
quietbits, line 1: Could we avoid using `any` in these? Just curious because they might be too much to type out.

common/lumens.js
quietbits: Do we need to keep this file in JavaScript?

acharb: yup, at least until the frontend uses typescript because it's importing this file
eg https://github.com/stellar/dashboard/blob/6ec7ac5bebc0d12f22a848a426ffdf1964f06e27/frontend/components/LumensCirculating.js#L5

backend/ledgers.ts
```diff
@@ -1,11 +1,18 @@
...
+interface Ledger {
+  date: string;
+  transaction_count: number;
+  operation_count: number;
+}
+
+let cachedData: Array<Ledger>;
 
-export const handler = function(req, res) {
+export const handler = function({}, res: Response) {
```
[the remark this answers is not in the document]
acharb: ya I wasn't quite sure for this, I found [this discussion](https://stackoverflow.com/a/50547932/8846296) which is why I went with `{}`. You don't have to add a `: any` and also `_` interferes with the `import _ from "lodash";` line

Is `_` the common way we're doing it in other ts projects? If we're using `_` most often I can change the import lodash line to something else

backend/ledgers.ts
```diff
@@ -20,19 +27,17 @@ function updateResults() {
...
 }
 
-function convertFields(ledger) {
-  // String to int fields
-  const fields = ["transaction_count", "operation_count"];
-  for (let field of fields) {
-    ledger[field] = parseInt(ledger[field]);
-  }
-
+function convertFields(ledger: any): Ledger {
```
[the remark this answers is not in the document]
acharb: this one's because the `transaction_count` and `operation_count` values when pulled from postgres come back as strings. I thought it might be too confusing if there was a second Ledger type with `transaction_count` and `operation_count` as type `string` instead of `number`, so I left it as `any`. wdyt? 🤔

common/lumens.d.ts
acharb: ah this file was auto-generated using the `tsc --declaration` command. Do you mean instead of using `any` using their correct types?

acharb: changed: https://github.com/stellar/dashboard/pull/204/commits/e4c7b25f9072d1aab99fe8cb301338af58669585
but could you still elaborate what you mean by `because they might be too much to type out.`, for my own understanding

backend/ledgers.ts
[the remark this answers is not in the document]
quietbits: I haven't seen this in our projects used often, we usually can omit the param which is not used. Either way is fine, I guess? I've seen `_` convention in other projects and suggestions, but as with everything in JS/TS world, it's not the only way. :sweat_smile:

backend/ledgers.ts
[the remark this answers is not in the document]
quietbits: Would it be OK to accept either `string` or a `number` for those props? That way we could use the same type in both places. Another option is to add an extra type that extends the main type. I personally like when everything is typed as much as possible, makes it easier to work with that code.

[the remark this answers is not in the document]
acharb: an extra type sgtm 👍  agreed less uses of `any` is better
https://github.com/stellar/dashboard/pull/204/commits/eece86d470600eda80ae0f9d7c066fe0cfaf1f7a

backend/ledgers.ts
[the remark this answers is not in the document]
acharb: changed to `_: any` , `{}` does look more ugly the more I look at it. Doesn't look like there's a single best way of handling (if only [this suggestion](https://github.com/microsoft/TypeScript/issues/29202#issuecomment-662060936) was accepted)
https://github.com/stellar/dashboard/pull/204/commits/483166a64bb025dcbbf4dd37137d092af2aaaa0f

backend/lumens.ts
```diff
@@ -1,9 +1,21 @@
...
+    ecosystemSupport: string;
+    useCaseInvestment: string;
+    userAcquisition: string;
+  };
+}
+
+export let cachedData: CachedData;
 
-export const v1Handler = function(req, res) {
+export const v1Handler = function({}, res: Response) {
```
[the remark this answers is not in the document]
acharb: https://github.com/stellar/dashboard/pull/204/commits/483166a64bb025dcbbf4dd37137d092af2aaaa0f

quietbits (approved):

Issue #179, opened 2021-10-22
`api/v2/lumens/` isn't built to be horizontally scalable


piyalbasu:
Currently, the endpoint `api/v2/lumens/` relies on cached data that is stored in memory. This creates an issue when we have multiple instance in kubernetes running the app. Depending on which image you hit, you'll get a slightly different payload as they (likely) all started running at different times and are running on different schedules.

Thoughts on a solution:
- Do we want to save cached data in a single database and use that as our source of truth? (The current postgresdb in the repo has the above problem of multiple instances)

Issue #198, opened 2021-12-07
Use Redis for caching


acharb:
- [x] Replace the in memory storage with Redis - will solve https://github.com/stellar/dashboard/issues/179

- [ ] Replace Postgres with Redis. Postgres is currently being used for caching anyways, and we don't foresee the schema getting more complicated, so using Redis only in this project makes sense.

PR #205, opened 2022-01-05
backend: connect to redis and replace in memory storage with Redis

acharb:
**WHAT**
connect to Redis and replace the in memory cachine with Redis

**WHY**
the in-memory caching has caused problems with horizontal scaling this product. See https://github.com/stellar/dashboard/issues/179. Replacing it with Redis will ensure api data returned will be the same regardless how many pods are up

note: Redis is now up on both prod and dev kube clusters: https://github.com/stellar/kube/pull/734

closes https://github.com/stellar/dashboard/issues/179
and fulfills first part of https://github.com/stellar/dashboard/issues/198


test/tests/unit/backend.ts
```diff
@@ -1,15 +1,21 @@
 import chai from "chai";
+import { redisClient } from "../../../backend/redis";
 const v1 = require("../../../backend/lumens");
 const v2v3 = require("../../../backend/v2v3/lumens");
 
 describe("lumens v1", function () {
+  this.timeout(10000);
```
acharb, line 7: these added test timeouts are not related to redis but were added bc occasionally calls to Horizon would take longer than usual (assuming traffic related) causing a timeout error (default is 2s). So 10s seemed to be a good timeout empirically

backend/ledgers.ts
```diff
@@ -16,10 +18,17 @@ interface LedgerSql {
...
-let cachedData: Array<Ledger>;
-
-export const handler = function (_: any, res: Response) {
-  res.send(cachedData);
+export const handler = async function (
+  _: any,
+  res: Response,
+  next: NextFunction,
+) {
+  let cached = await redisClient.get("ledgers");
```
piyalbasu, line 26: since this var isn't reassigned, you should use `const` here. 

I see that we don't have eslint setup for this repo yet. Our @stellar/eslint-config will pick up this error for you. No need to implement eslint on this PR as it will likely complain about a lot of nit picky stuff, but probably a good idea to implement sooner rather than later. 

Example: https://github.com/stellar/account-viewer-v2/blob/master/.eslintrc.js

backend/lumens.ts
```diff
@@ -13,10 +14,22 @@ interface CachedData {
...
+export const v1Handler = async function (
+  _: any,
+  res: Response,
+  next: NextFunction,
+) {
+  try {
+    let cachedData = await redisClient.get("lumensV1");
+    if (cachedData == null) {
+      return next(Error("null value found"));
+    }
```
piyalbasu, line 26: I see that for a lot of these handlers you're following a pattern of checking `cachedData` and returning an error if it's null. Would it be helpful to create a helper that handles this so you don't have to keep doing this check manually?

test/tests/unit/backend.ts
piyalbasu: Do you mind adding a comment at the top of this file that explains this?

piyalbasu (approved):

test/tests/unit/backend.ts
acharb: 👍 https://github.com/stellar/dashboard/pull/205/commits/7beea1a978ada649f6bf18bc0d0bd17388269079

backend/ledgers.ts
acharb: https://github.com/stellar/dashboard/pull/205/commits/15aa33be37824ac00766761c729086d0c061f355

good idea, I'll add it next, tracked here: https://github.com/stellar/dashboard/issues/207

backend/lumens.ts
acharb: sgtm, I went back and forth originally but leaned towards [writing code that's easy to delete over easy to extend](https://github.com/stellar/dashboard/pull/205/commits/62b6f5890035aef35e5d8380804114b532a962b3) (an article I got from @leighmcculloch  😄 ). Looking at it again it does look cleaner with a helper though, added 👍 
https://github.com/stellar/dashboard/pull/205/commits/62b6f5890035aef35e5d8380804114b532a962b3

PR #209, opened 2022-01-12
backend: remove postgres, use redis for api/ledgers data

acharb:
**WHAT**
remove use of postgres on the backend, and use redis for the `api/ledgers/public` data

**WHY**
The dashboard api endpoints essentially cache organized data given from Horizon, which is more of a use case for redis than postgres. And since postgres was just being used for a single purpose (storing all ledgers), we can now remove it and have a single storage option on the backend, reducing the complexity of the backend. In order to do this the data is now organized before being stored in the cache, as opposed to after like it was with postgres.

closes https://github.com/stellar/dashboard/issues/198
part of the backend refactor [plan](https://stellarorg.atlassian.net/wiki/spaces/PET/pages/4244078593/Dashboard+Backend+Upgrade+Plan)


backend/app.ts
```diff
@@ -3,8 +3,17 @@
...
-// Run backend with interval cache updates.
+// Run backend with cache updates.
 const { updateLumensCache } = require("./routes");
+const { updateLedgers } = require("./ledgers");
 
-setInterval(updateLumensCache, 10 * 60 * 1000);
-updateLumensCache();
+(async () => {
+  setInterval(updateLumensCache, 10 * 60 * 1000);
+  console.log("starting lumens cache update");
```
quietbits, line 12: Do we want to keep these `console.log` here and below?

backend/ledgers.ts
```diff
@@ -12,94 +10,124 @@ interface Ledger {
...
-
-  postgres.sequelize
-    .query(query, { type: QueryTypes.SELECT })
-    .then((results: Array<LedgerSql>) => {
-      const cachedData: Array<Ledger> = map(results, convertFields);
-      redisClient.set("ledgers", JSON.stringify(cachedData));
+export async function updateLedgers() {
+  let pagingToken = await redisClient.get("paging_token");
+  if (pagingToken == null || pagingToken == "") {
+    pagingToken = "now";
```
quietbits, line 36: We could make "now" string a const as well to stay consistent with other string consts.

backend/ledgers.ts
```diff
@@ -12,94 +10,124 @@ interface Ledger {
...
+  ledgersKey: string,
+  pagingTokenKey: string,
+) {
+  if (!ledgers.length) {
+    console.log("no ledgers to update");
+    return;
   }
-});
+  const json = (await redisClient.get(ledgersKey)) || "[]";
+  let cachedLedgers: Array<Ledger> = JSON.parse(json);
```
quietbits, line 86: Could be `const`

leighmcculloch (approved):
I have a few questions, I think driven by just lack of knowledge on my part on how this works. These questions don't need to block anything, except maybe the question on how ledger data is being added together, I didn't expect that.

backend/ledgers.ts
```diff
@@ -12,94 +10,124 @@ interface Ledger {
...
+  let pagingToken = "";
+
+  ledgers.forEach((ledger: LedgerRecord) => {
+    let date = ledger.closed_at.substring(5, 10);
```
quietbits, line 90: Looks like this could be a `const` as well.

backend/ledgers.ts
```diff
@@ -12,94 +10,124 @@ interface Ledger {
...
+    let index = findIndex(cachedLedgers, { date: date });
```
quietbits, line 91: We could write just `{date}`. In modern JS and TS, if the property name and variable are the same, providing only property works. :nerd_face: 

Also, index can be a `const`

test/tests/unit/backend.ts
```diff
@@ -1,8 +1,12 @@
 import chai from "chai";
 import { redisClient } from "../../../backend/redis";
+import { updateCache, catchup, LedgerRecord } from "../../../backend/ledgers";
 const v1 = require("../../../backend/lumens");
 const v2v3 = require("../../../backend/v2v3/lumens");
 
+const REDIS_LEDGER_KEY = "ledgers_test";
+const REDIS_PAGING_TOKEN_KEY = "paging_token_test";
```
quietbits, line 8: These were defined earlier in the code, maybe we could move them to a separate file (`constants.ts`, for example)? That way we can import them where needed.

test/tests/unit/backend.ts
```diff
@@ -126,3 +130,132 @@ describe("lumens v3", function () {
...
+          successful_transaction_count: 30,
+          failed_transaction_count: 20,
+          operation_count: 200,
+          closed_at: "2022-01-12T01:07:00Z",
+        },
+      ];
+
+      await updateCache(ledgers, REDIS_LEDGER_KEY, REDIS_PAGING_TOKEN_KEY);
+
+      let cachedLedgers = await redisClient.get(REDIS_LEDGER_KEY);
```
quietbits, line 170: We might replace all of these `let`s with `const` in this file.

backend/ledgers.ts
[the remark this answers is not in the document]
acharb: the ledger `closed_at` comes from the ledger in this format: `"2022-01-13T20:54:32Z"`

using `.substring` was how it was used before, but I just noticed it also used the postgres sql query formatting, so probably more brittle than I thought now. I added a helper for better formatting:
https://github.com/stellar/dashboard/pull/209/commits/ccc4c751b63c64c604b9a8027bdb41610cedf342

backend/ledgers.ts
```diff
@@ -12,94 +10,124 @@ interface Ledger {
...
+      });
+    } else {
+      cachedLedgers.splice(index, 1, {
+        date: date,
+        transaction_count:
+          cachedLedgers[index].transaction_count +
+          ledger.successful_transaction_count +
+          ledger.failed_transaction_count,
+        operation_count:
+          cachedLedgers[index].operation_count + ledger.operation_count,
```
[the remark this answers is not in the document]
acharb: maybe the naming should be changed. The cache is stored like so:
```
[
  {
    "date": "01-13",
    "transaction_count": 5245946,
    "operation_count": 10088417
  },
  {
    "date": "01-12",
    "transaction_count": 5787617,
    "operation_count": 11333624
  },
...
```
which is what `cachedLedgers` is referring to and is what's being updated. Which technically isn't a ledger, I couldn't think of a better word 🤔 . This code no longer stores every `LedgerRecord` from horizon like it did before with postgres.

I think maybe this would be clearer when LedgerRecord is not `any` anymore https://github.com/stellar/dashboard/blob/ccc4c751b63c64c604b9a8027bdb41610cedf342/backend/ledgers.ts#L13-L14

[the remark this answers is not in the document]
acharb: `LedgerStat`? this was referring to something else with postgres before I removed it, but I think could be fitting here. `LedgerAggregate` might be better, but a mouth full. Thoughts on either of those?

backend/app.ts
acharb: i'd like to ya , when looking at the kube logs it helps with debugging/monitoring to see when the updates are started (since not at the same time now due to Horizon rate limiting)

backend/ledgers.ts
```diff
@@ -12,94 +10,124 @@ interface Ledger {
...
-    group by to_char(closed_at, 'YYYY-MM-DD')
-    order by 1 desc`;
-
-  postgres.sequelize
-    .query(query, { type: QueryTypes.SELECT })
-    .then((results: Array<LedgerSql>) => {
-      const cachedData: Array<Ledger> = map(results, convertFields);
-      redisClient.set("ledgers", JSON.stringify(cachedData));
+export async function updateLedgers() {
+  let pagingToken = await redisClient.get("paging_token");
```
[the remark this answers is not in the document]
acharb: good 👁️ 
https://github.com/stellar/dashboard/pull/209/commits/3aa116a54534ef1f93a1d32d26964f38c83c04dd

backend/ledgers.ts
acharb: https://github.com/stellar/dashboard/pull/209/commits/3aa116a54534ef1f93a1d32d26964f38c83c04dd

backend/ledgers.ts
```diff
@@ -12,94 +10,124 @@ interface Ledger {
...
+  pagingToken: string,
+  pagingTokenKey: string,
+  limit: number, // if 0, catchup until now
+) {
+  const horizon = new stellarSdk.Server("https://horizon.stellar.org");
+  let ledgers = [];
+  let total = 0;
+
+  while (true) {
+    let resp = await horizon.ledgers().cursor(pagingToken).limit(200).call();
```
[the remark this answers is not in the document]
acharb: https://github.com/stellar/dashboard/pull/209/commits/3668f052cf5b80466181349192ea2a67ed6e1e36

backend/ledgers.ts
acharb: https://github.com/stellar/dashboard/pull/209/commits/3668f052cf5b80466181349192ea2a67ed6e1e36

backend/ledgers.ts
acharb: https://github.com/stellar/dashboard/pull/209/commits/3668f052cf5b80466181349192ea2a67ed6e1e36

test/tests/unit/backend.ts
acharb: https://github.com/stellar/dashboard/pull/209/commits/3668f052cf5b80466181349192ea2a67ed6e1e36

thx for catching, if I missed any my next PR adding eslint should pick them up

backend/ledgers.ts
acharb: good to know 👍 
https://github.com/stellar/dashboard/pull/209/commits/611c5a04d139e54bfb6dee3ebf9cf78ff59e6c9d

test/tests/unit/backend.ts
acharb: 🤔  I'd like to keep these in the test file for now since they're constants specific to these tests. Maybe if they're used more widely for tests we could add to a different file. But the naming could probably be better, changed to make it clear used for the redis testing only: https://github.com/stellar/dashboard/pull/209/commits/85e319459c7a103a361983cd1c758de1540c9524

backend/ledgers.ts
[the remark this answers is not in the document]
acharb: I like `LedgerStat` the more I think of it: https://github.com/stellar/dashboard/pull/209/commits/e6fb860861857a5c4479e4bc9b15c396c99b7e12

if you disagree lmk though

Issue #207, opened 2022-01-10
all: add eslint


acharb:
use `@stellar/eslint-config` so the code is aligned with other SDF projects.

eg. https://github.com/stellar/account-viewer-v2/blob/master/.eslintrc.js

PR #211, opened 2022-01-15
all: add eslint

acharb:
**WHAT**
adding eslint to the project, and running as a pre-commit hook on backend code changes

closes https://github.com/stellar/dashboard/issues/207

**WHY**
linting is good practice for js/ts projects to prevent runtime errors, that `tsc` or `prettier` wouldn't catch


package.json
```diff
@@ -13,23 +13,19 @@
...
+    "pre-commit": "lint-staged"
   },
   "lint-staged": {
     "**/*.js": [
-      "prettier --write",
-      "git add"
-    ]
+      "prettier --write"
+    ],
+    "./backend/*.ts": "eslint --cache"
```
acharb, line 24: only adding to backend for now. During frontend refactor we can include it, the eslint rules may need altering.

piyalbasu (approved):
Awesome!

charlestonXdao (commented):
good

PR #232, opened 2022-02-14
v2: Network status

quietbits:
This PR also includes:
- Redux store setup
- Ledger data to be used in Ledger sections

The protocol update date is hardcoded, will need to add a link as well.

image

### DIFF
--- a/src/frontend/components/NetworkStatus/index.tsx
+++ b/src/frontend/components/NetworkStatus/index.tsx
@@ -1,5 +1,137 @@
-import { Card } from "@stellar/design-system";
+import { useState, useEffect } from "react";
+import { Card, TextLink, Icon, Tag } from "@stellar/design-system";
+import { useDispatch } from "react-redux";
+import moment from "moment";
+
+import {
+  fetchLedgersAction,
+  startLedgerStreamingAction,
+  stopLedgerStreamingAction,
+} from "frontend/ducks/ledgers";
+import { useRedux } from "frontend/hooks/useRedux";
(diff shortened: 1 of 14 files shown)


src/frontend/pages/Testnet.tsx
```diff
@@ -1 +1,8 @@
-export const Testnet = () => <div>Testnet</div>;
+import { NetworkStatus } from "frontend/components/NetworkStatus";
+import { Network } from "types";
+
+export const Testnet = () => (
+  <div className="ContentWrapper">
+    <NetworkStatus network={Network.TESTNET} />
```
quietbits, line 6: I thought it will be more efficient to pass the testnet prop manually to components instead of detecting it dynamically on every component (default network is mainnet, so no need to pass it there). We already know the network where the components are used.

src/frontend/components/NetworkStatus/index.tsx
```diff
@@ -1,5 +1,137 @@
-import { Card } from "@stellar/design-system";
+import { useState, useEffect } from "react";
+import { Card, TextLink, Icon, Tag } from "@stellar/design-system";
+import { useDispatch } from "react-redux";
+import moment from "moment";
```
piyalbasu, line 4: The Moment dev's generally advocate for not using the lib anymore (esp in new projects). 
https://momentjs.com/docs/

It might be best to not use it in new v2 code and start coming up with a plan to refactor old instances of it

piyalbasu (approved):

PR #237, opened 2022-02-15
v2: Last 10 ledgers info

quietbits:
- Moved `LedgerClosedTime` to a separate component

image

### DIFF
--- a/src/frontend/ducks/ledgers.ts
+++ b/src/frontend/ducks/ledgers.ts
@@ -102,6 +102,11 @@ export const startLedgerStreamingAction = createAsyncThunk<
         .stream({
           onmessage: (ledger: LedgerRecord) => {
             const { lastLedgerRecords } = ledgersSelector(getState());
+
+            if (!lastLedgerRecords[0]) {
+              return;
+            }
+
             const timeDiff = getDateDiffSeconds(
               ledger.closed_at,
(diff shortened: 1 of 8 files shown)


src/frontend/ducks/ledgers.ts
```diff
@@ -102,6 +102,11 @@ export const startLedgerStreamingAction = createAsyncThunk<
         .stream({
           onmessage: (ledger: LedgerRecord) => {
             const { lastLedgerRecords } = ledgersSelector(getState());
+
+            if (!lastLedgerRecords[0]) {
```
piyalbasu, line 106: just curious: is it possible for this array to have a length but have no record (or null) at index 0? I'm assuming no

piyalbasu: mainly wondering why you chose to check for this instead of `lastLedgerRecords.length`

piyalbasu (approved):

src/frontend/ducks/ledgers.ts
quietbits: Good point, we could be checking just for the array length here, but I went with the `lastLedgerRecords[0]` because it's used below in `lastLedgerRecords[0].closedAt`. Do you think it would more clear to check for the array length?

PR #238, opened 2022-02-16
v2: Replace and remove moment.js

quietbits:
Using native JavaScript where possible and `date-fns` where it's more convenient.


piyalbasu (approved):
Nice!! Thanks for doing this

PR #243, opened 2022-02-25
v2: refactor code structure

quietbits:
Frontend and backend are built separately now. Most changes are for moving files around + configs.

I couldn't get `lint-staged` working, so leaving that for later.


package.json
```diff
@@ -14,16 +14,19 @@
...
+    "start:backend": "cd backend && yarn start && cd ..",
+    "clean:frontend": "cd frontend && yarn clean && cd ..",
+    "clean:backend": "rm -rf node_modules && rm -rf build && cd backend && yarn clean && cd ..",
+    "clean": "run-p clean:frontend clean:backend",
+    "prebuild:frontend": "cd frontend && yarn install && cd ..",
+    "prebuild:backend": "cd backend && yarn install && cd ..",
+    "prebuild": "run-s prebuild:backend prebuild:frontend",
+    "build:frontend": "cd frontend && yarn build && cd ..",
+    "build:backend": "cd backend && yarn build && cd ..",
+    "build": "run-s build:frontend build:backend"
```
piyalbasu, line 29: is there a reason we're not using `workspaces`? It would simplify a lot of these build steps

piyalbasu (approved):

acharb (approved):
lgtm!

only thing is the backend tests aren't part of the workflow anymore, and the yarn test command was removed. Not critical, but we'd probably want to include that (maybe after all your v2 frontend changes so doesn't slow down your workflow) to be safe with next backend changes. I added a ticket to track https://github.com/stellar/dashboard/issues/245

package.json
quietbits: It took us a while to figure out how to build frontend and backend properly. Adding an extra variable in config seemed a bit too much. :sweat_smile: We should definitely try workspaces down the line. :100:

quietbits:
@acharb, thanks for opening an issue for tests. We'll definitely fix them later. I had weird timing issues running tests with GitHub Actions, but new build process might have fixed it. :crossed_fingers:

PR #261, opened 2022-05-06
SDFD-3_add_tests

Selhar:
- cherry picked [public API commit](https://github.com/stellar/dashboard/commit/17cbc05cb3df4a7878b72d07c915e37fac772b2c) from master into this branch
- Since we were already using ts-node instead of babel, i made use of ts-jest as our transpiler. The idea is not to deviate from current standards.
- Moved gcloud folder into the backend domain, since it hasn't been used anywhere else. Let me know if the location is not ideal @acharb 
- Timeout to fetch data has been greatly increased. From local testing, it could take between 30 to 4x seconds to fetch redis data, so i left it at 50 seconds. It might be due to my gcloud account, but either way it would be ideal to have a higher timeout in case somebody else picks up the project, since the errors on timeout are not very intuitive.


acharb (approved):
lgtm!

acharb:
looks like the test script also needs to be added https://github.com/stellar/dashboard/blob/17cbc05cb3df4a7878b72d07c915e37fac772b2c/package.json#L15

Selhar:
Oh, the issue is that we're running the script from the root folder. Can i change it to the backend folder so that we use the correct package.json? Are there any front end tests that should also run during pull requests/merges? I can make a separate action for them if that's the case. 
@acharb @quietbits

acharb:
@Selhar  adding it to the backend package.json sgtm 👍  the frontend doesn't have any tests atm (I don't think we have plans to add any either)

Selhar:
Fixed! The default directory for this action is now `./backend`. I also changed the job title to more clearly reflect what is being tested.

backend/src/bigQuery.ts
```diff
@@ -1,21 +1,30 @@
...
-};
+let options;
+if (process.env.DEV) {
+  options = {
+    keyFilename: "gcloud/service-account.json",
+    projectId: process.env.BQ_PROJECT_ID,
+  };
+} else {
+  options = {
+    keyFilename: "../gcloud/service-account.json",
```
acharb, line 11: this may not be the right location of the `service-account.json` in production before it was `keyFilename: "../../../gcloud/service-account.json"`, but we can check in the preview environment before merging

acharb: oh it doesn't look like previews show up on forks, let me create a branch on the dashboard repo with these changes to check

Selhar: The tests have been running for an awfully long time, the only noticeable difference i could notice is that our linter is changing the health checks on redis, i'll push a modification to circumvent that. Current workflow can be stopped though, it seems to be on an infinite loop.

Selhar: I pushed the new commit, it adds our .github folder to the prettierignore list. It's a long shot as to what may be causing the loop, but github actions are notoriously iffy about formatting so it may be it. I'll keep an eye on the next workflow to get the logs.

acharb: also let's change the gcloud prod file location back to `keyFilename: "../../../gcloud/service-account.json",`. Doesn't look like the preview environment pod can find the file currently and this was what it was set to previously that I'm fairly certain was working

Selhar: Done! ammended the previous commit.

Selhar: I got the logs this time, this recent test also failed and will run indefinitely. I'll investigate, meanwhile the last workflow can be cancelled. Apparently the packages are not being installed.

Selhar: Pushing a new commit with the following changes:
- Reintroduced a forced exit after all tests run. Dangling redis instances were causing the forever loop.
- Added a new step to our github action to install the dependencies of our commons scripts. Since lumens makes use of src/commons/lumens.js and that requires a different set of packages, they have to be installed separately.

Hopefully the tests will run without issue after these changes.

acharb: nice catches 👍

acharb: nice catches 👍

acharb (approved):
lgtm!

btw you guys have been added as contributors to this project now so you should be able to create branches and see previews for your PRs now

backend/src/bigQuery.ts
Selhar: whats the merge policy @acharb ? can i merge once it's approved or is there another proccess?

acharb: yup after approved you can merge 👍

PR #268, opened 2022-05-11
SDFD-6 Feat: By the numbers section

### LINKED ISSUE #255: By the numbers
Build the "By the numbers" section following the design.
- The container component `ByTheNumbers` is already [created](https://github.com/stellar/dashboard/tree/dashboard-v2/frontend/src/components/ByTheNumbers).
- Use [Stellar Design System](https://design-system.stellar.org/) (SDS) components as much as possible.
- Use only colors from the SDS to make sure light and dark modes "just work". We don't have color palette on the SDS website yet, so they can be found [here](https://github.com/stellar/stellar-design-system/blob/main/%40stellar/design-system/src/global.scss) or viewed in browser dev tools.
- Icons should also be used from the SDS (`<Icon />` component). It's OK if they differ slightly.
- Will need a backend endpoint to get the data.
- Only on mainnet (no testnet).

image

image

Some notes

image

italomlp:
Update the By the numbers section using DEX data

<details>
<summary>Screenshot:</summary>

image
PS: Arbitrary values
</details>

Closes #255

### DIFF
--- a/frontend/src/components/AmountInfoCard/index.tsx
+++ b/frontend/src/components/AmountInfoCard/index.tsx
@@ -1,0 +1,45 @@
+import { ReactNode } from "react";
+import { Card, Heading5 } from "@stellar/design-system";
+
+import "./styles.scss";
+
+type Props = {
+  title: string;
+  titleIcon?: ReactNode;
+  amount: string | number;
+  amountBy?: string | number;
+  amountPrefixContent?: ReactNode;
+  description?: string;
+};
+
+export const AmountInfoCard = ({
+  title,
+  titleIcon,
+  amount,
+  amountBy,
+  amountPrefixContent,
+  description,
+}: Props) => (
+  <div className="AmountInfoCard">
+    <Card noShadow>
+      <div className="AmountInfoCard__card_title">
+        {titleIcon}
+        <div className="AmountInfoCard__card_title__title">{title}</div>
+      </div>
+      <div className="AmountInfoCard__content">
+        {amountPrefixContent}
+        <Heading5>
+          {amount}
+          {amountBy && (
+            <span className="AmountInfoCard__content__amount_by">
+              {amountBy}
+            </span>
+          )}
+        </Heading5>
+      </div>
+      {description && (
+        <div className="AmountInfoCard__description">{description}</div>
+      )}
+    </Card>
+  </div>
+);
--- a/frontend/src/components/AmountInfoCard/index.tsx
+++ b/frontend/src/components/AmountInfoCard/index.tsx
@@ -23,21 +23,23 @@ export const AmountInfoCard = ({
   <div className="AmountInfoCard">
     <Card noShadow>
       <div className="AmountInfoCard__card_title">
-        {titleIcon}
+        {!!titleIcon && (
+          <div className="AmountInfoCard__card_title__icon">{titleIcon}</div>
+        )}
         <div className="AmountInfoCard__card_title__title">{title}</div>
       </div>
       <div className="AmountInfoCard__content">
         {amountPrefixContent}
         <Heading5>
           {amount}
-          {amountBy && (
+          {!!amountBy && (
             <span className="AmountInfoCard__content__amount_by">
               {amountBy}
             </span>
           )}
         </Heading5>
       </div>
-      {description && (
+      {!!description && (
         <div className="AmountInfoCard__description">{description}</div>
       )}
     </Card>
--- a/frontend/src/components/AmountInfoCard/styles.scss
+++ b/frontend/src/components/AmountInfoCard/styles.scss
@@ -1,0 +1,57 @@
+.AmountInfoCard {
+  flex: 1;
+  min-width: 150px;
+
+  .Card {
+    padding: 1rem;
+    display: flex;
+    flex-direction: column;
+    justify-content: space-between;
+  }
+
+  &__card_title {
+    color: var(--pal-text-tertiary);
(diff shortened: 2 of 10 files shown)


Selhar:
Hey @acharb, we're getting failed requests on the preview environment, do we have any sort of logs to check out what might be happening?

frontend/src/components/AmountInfoCard/index.tsx
```diff
@@ -0,0 +1,45 @@
...
+      </div>
+      <div className="AmountInfoCard__content">
+        {amountPrefixContent}
+        <Heading5>
+          {amount}
+          {amountBy && (
+            <span className="AmountInfoCard__content__amount_by">
+              {amountBy}
+            </span>
+          )}
```
quietbits, line 37: Just to be safe, we prefer to use a conditional operator in cases like this (`amountBy ? <span /> : null`). Even though there should be no issues in this specific scenario, it's good to stay consistent. Same for `description` below.

frontend/src/components/AmountInfoCard/styles.scss
```diff
@@ -0,0 +1,57 @@
...
+      font-weight: var(--font-weight-medium);
+
+      &:not(:first-child) {
+        margin-left: 0.4rem;
+      }
+    }
+
+    &:not(:last-child) {
+      margin-bottom: 0.4rem;
+    }
```
quietbits, line 44: Maybe this could be achieved with `gap` in `flex` parent?

frontend/src/components/ByTheNumbers/index.tsx
```diff
@@ -1,14 +1,82 @@
...
+
+  return (
+    <SectionCard
+      title="By the Numbers"
+      titleLinkLabel="API"
+      titleLink="/api/dex/all"
+      isLoading={dex.status === ActionStatus.PENDING}
+      noData={!dexData}
+    >
+      {dexData && (
```
quietbits, line 50: Same here about the conditional operator.

frontend/src/components/ByTheNumbers/index.tsx
```diff
@@ -1,14 +1,82 @@
...
+              icon={<Icon.RefreshCcw width={16} height={16} />}
+              last24HRValue={dexData.trades.last24HR}
+              overallValue={dexData.trades.overall}
+              fluctuation={dexData.trades.fluctuation}
+            />
+          </div>
+          <div className="ByTheNumbers__amount_cards_container">
+            <AmountInfoCard
+              title="Total Unique Assets"
+              titleIcon={<Icon.Database width={16} height={16} />}
```
quietbits, line 64: The API might be simpler if the component handled icon size in CSS. That way user doesn't need to remember to set the size on the icon here. What do you think?

frontend/src/components/DexCard/index.tsx
```diff
@@ -0,0 +1,58 @@
+import { ReactNode, useMemo } from "react";
+import { Card, Heading4, Tag, Eyebrow } from "@stellar/design-system";
+
+import "./styles.scss";
+import BigNumber from "bignumber.js";
```
quietbits, line 5: NIT: relative imports should be at the end.

quietbits (approved):
Looks nice! :tada: Just left a few optional comments, but nothing blocking.

acharb:
> Hey @acharb, we're getting failed requests on the preview environment, do we have any sort of logs to check out what might be happening?

@Selhar unfortunately the preview logs are internal atm, I can ask our ops team if there's a way you guys can view but I have a feeling it will be no :/

but I see the failed request, if you make the failing request directly in your browser: https://dashboard-pr268.prototypes.kube001.services.stellar-ops.com/api/dex/all you'll see the error returned is:
```
TypeError: Cannot read properties of null (reading '0')
    at /app/backend/build/dex/data.js:88:43
    at Generator.next (<anonymous>)
    at fulfilled (/app/backend/build/dex/data.js:5:58)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
```

Selhar:
Oh, the google cloud key is not available in the preview environment. @italomlp had this issue before i added the key to his .env.
Is this something we can fix? @acharb

acharb:
@Selhar Yup good find. We probably could add the key to the preview pods, but for now I manually added the data to redis (this should persist since we're not clearing that atm). I think better (at least for now) to not  add the gcloud key to all the preview pods to limit BQ requests since there can be a lot of previews running at once.  Regardless the dex backend calls work now for previews 👍

frontend/src/components/AmountInfoCard/index.tsx
italomlp: I agree with you. In these cases, I usually add a "double negative" before the check expression so that it always will be converted to a boolean value, which in JSX renders nothing, meaning that we keep consistency. What do you think about this approach? It would be something like:
```suggestion
          {!!amountBy && (
            <span className="AmountInfoCard__content__amount_by">
              {amountBy}
            </span>
          )}
```

frontend/src/components/ByTheNumbers/index.tsx
italomlp: Yeah, that makes sense, and we can use `rem` instead of pixels. 🚀

frontend/src/components/AmountInfoCard/index.tsx
quietbits: Sure, that works. :+1:

quietbits (approved):

PR #270, opened 2022-05-12
SDFD-19 Feat: Add missing charts for existing sections

### LINKED ISSUE #252: Charts
Find a charting library or libraries (we have simple as well as more detailed charts) that we can use. Main criteria:
1. It needs to match as close to our designs as possible (if there is something really good and can't be customized to look the way we want, we can still consider it).
2. Styled with CSS. We need it to work in light and dark mode easily, so it would be great if we could style charts using CSS variables we already have. A CSS-in-JS solution might work, but pure CSS is preferable.
3. It would be great if each chart would become a standalone, reusable component that we could later add to the Stellar Design System.

Charts we have on dashboard

image

image

image

image

image

image

italomlp:
## About

Add missing charts for existing sections of Dashboard:
- Last Ledger
- Last 10 Ledgers info
- Total XLM Issued
- Network nodes

Closes #252 

## Screenshots

<details>
<summary>Some screenshots for illustration:</summary>

All these charts are also working on Dark mode and for Testnet (if applicable). Available on PR preview.

Screen Shot 2022-05-12 at 17 30 50
Screen Shot 2022-05-12 at 17 31 01
Screen Shot 2022-05-12 at 17 31 38
Screen Shot 2022-05-12 at 17 31 11

</details>

### DIFF
--- a/frontend/src/components/LastLedger/index.tsx
+++ b/frontend/src/components/LastLedger/index.tsx
@@ -1,8 +1,11 @@
 import { Icon } from "@stellar/design-system";
+
 import { networkConfig } from "constants/settings";
 import { useRedux } from "hooks/useRedux";
 import { SectionCard } from "components/SectionCard";
 import { LedgerClosedTime } from "components/LedgerClosedTime";
+import { CircularChart } from "components/charts/CircularChart";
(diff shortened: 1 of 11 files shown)


quietbits (approved):
Charts look great! :tada:

PR #272, opened 2022-05-23
SDFD-6_dex_data

Selhar:
Hey, this will be a huge PR, sorry for the massive amounts of text, there are too many details and i'd rather have it all in one place to make it easier to catch any mistakes.

## New endpoints

GET `/api/v2/dex/24h-payments`:  Integer
GET `/api/v2/dex/24h-trades`: Object:
```
{
  trades_last_24h: Integer,
  change: String,
  overall: Integer
}
``` 
GET `/api/v2/dex/unique-assets`: Integer
GET `/api/v2/dex/active-accounts` : Integer
GET `/api/v2/dex/volume`: Object: 
```
{
  volume: Integer,
  change: String,
  overall: Integer
}
```
GET `/api/v2/dex/all`: Object: 
```
{
  volume: volume endpoint object
  uniqueAssets: integer
  tradesLast24h: trades endpoint object
  uniqueAssets: integer
  activeAccounts: integer
}
```

Example output (volume is completely wrong because we can't get a price for all assets):
image

### Get asset price
```https://api.stellar.expert/explorer/public/asset/${asset}/stats-history```
### Get volume
This one is very.. very tricky. There are 3 queries here, one is supposed to fetch the results for the last 24h, another for the 48h and a third one for the overall data. All of these queries return the same amount of rows, 27262. I'm not sure if i'm missing something but it's as if the filter is being ignored completely. Here are the queries:

**24h:**
```
      WITH bought_trades AS
    (
        SELECT SUM(buying_amount) AS summed_amount, CONCAT(buying_asset_code, "-", buying_asset_issuer) AS asset_name,
        FROM crypto-stellar.crypto_stellar_2.history_trades
        WHERE selling_liquidity_pool_id IS NULL OR selling_liquidity_pool_id = '' 
        
    AND ledger_closed_at >= TIMESTAMP_ADD(CURRENT_TIMESTAMP(), INTERVAL -24 HOUR)
    AND batch_run_date >= DATE_SUB(CURRENT_DATE(), INTERVAL 2 DAY) 
  
        GROUP BY asset_name
            UNION ALL
        SELECT SUM(buying_amount) AS summed_amount, "native" as asset_name
        from crypto-stellar.crypto_stellar_2.history_trades
        WHERE selling_liquidity_pool_id IS NULL OR selling_liquidity_pool_id = '' 
        AND buying_asset_TYPE = "native" 
        
    AND ledger_closed_at >= TIMESTAMP_ADD(CURRENT_TIMESTAMP(), INTERVAL -24 HOUR)
    AND batch_run_date >= DATE_SUB(CURRENT_DATE(), INTERVAL 2 DAY) 
  
        GROUP BY buying_asset_type
    )
    SELECT * FROM bought_trades;
```
**48h:**
```
      WITH bought_trades AS
    (
        SELECT SUM(buying_amount) AS summed_amount, CONCAT(buying_asset_code, "-", buying_asset_issuer) AS asset_name,
        FROM crypto-stellar.crypto_stellar_2.history_trades
        WHERE selling_liquidity_pool_id IS NULL OR selling_liquidity_pool_id = '' 
        
    AND ledger_closed_at >= TIMESTAMP_ADD(CURRENT_TIMESTAMP(), INTERVAL -48 HOUR)
    AND batch_run_date >= DATE_SUB(CURRENT_DATE(), INTERVAL 3 DAY) 
  
        GROUP BY asset_name
            UNION ALL
        SELECT SUM(buying_amount) AS summed_amount, "native" as asset_name
        from crypto-stellar.crypto_stellar_2.history_trades
        WHERE selling_liquidity_pool_id IS NULL OR selling_liquidity_pool_id = '' 
        AND buying_asset_TYPE = "native" 
        
    AND ledger_closed_at >= TIMESTAMP_ADD(CURRENT_TIMESTAMP(), INTERVAL -48 HOUR)
    AND batch_run_date >= DATE_SUB(CURRENT_DATE(), INTERVAL 3 DAY) 
  
        GROUP BY buying_asset_type
    )
    SELECT * FROM bought_trades;
```

**Overall:**
```
      WITH bought_trades AS
    (
        SELECT SUM(buying_amount) AS summed_amount, CONCAT(buying_asset_code, "-", buying_asset_issuer) AS asset_name,
        FROM crypto-stellar.crypto_stellar_2.history_trades
        WHERE selling_liquidity_pool_id IS NULL OR selling_liquidity_pool_id = '' 
        
        GROUP BY asset_name
            UNION ALL
        SELECT SUM(buying_amount) AS summed_amount, "native" as asset_name
        from crypto-stellar.crypto_stellar_2.history_trades
        WHERE selling_liquidity_pool_id IS NULL OR selling_liquidity_pool_id = '' 
        AND buying_asset_TYPE = "native" 
        
        GROUP BY buying_asset_type
    )
    SELECT * FROM bought_trades;
```

By the way, for each of these queries i must call the stellar expert endpoint to get the price for each row. So right now there are 27262 * 3 stellar expert requests (after a while i just get blocked from making them, so the code is currently limiting it to 100 queries max, i left a TODO in there while we figure it out).

### Trades query
**24h:**
```
    SELECT count(*) as data 
    FROM crypto-stellar.crypto_stellar_2.history_trades
    
    WHERE ledger_closed_at >= TIMESTAMP_ADD(CURRENT_TIMESTAMP(), INTERVAL -24 HOUR) 
    AND batch_run_date >= DATE_SUB(CURRENT_DATE(), INTERVAL 2 DAY)
```

**48h:**
``` 
    SELECT count(*) as data 
    FROM crypto-stellar.crypto_stellar_2.history_trades
    
    WHERE ledger_closed_at >= TIMESTAMP_ADD(CURRENT_TIMESTAMP(), INTERVAL -48 HOUR) 
    AND batch_run_date >= DATE_SUB(CURRENT_DATE(), INTERVAL 3 DAY)
```

**Overall:**
```
    SELECT count(*) as data 
    FROM crypto-stellar.crypto_stellar_2.history_trades
```

### Payments
```
    SELECT count(*) as data 
    FROM crypto-stellar.crypto_stellar_2.enriched_history_operations 
    WHERE type in (1, 2, 13) 
    AND closed_at >= TIMESTAMP_ADD(CURRENT_TIMESTAMP(), INTERVAL -24 HOUR) 
    AND successful is true
```

### Unique assets
```
    with all_assets as (
      select asset_code, asset_issuer, asset_type
      from crypto-stellar.crypto_stellar_2.history_assets
      group by asset_code, asset_issuer, asset_type
    )
    select count(distinct concat(aa.asset_code, ":", aa.asset_issuer)) as data
    from all_assets aa
    left outer join crypto-stellar.crypto_stellar_2.trust_lines tl
      on tl.asset_code = aa.asset_code
      and tl.asset_issuer = aa.asset_issuer
    where tl.asset_code is not null;
```

### Active accounts
```
    SELECT COUNT(DISTINCT(op_source_account))
    FROM crypto-stellar.crypto_stellar_2.enriched_history_operations 
    where type in (1, 2, 6, 13)
    AND closed_at >= TIMESTAMP_ADD(CURRENT_TIMESTAMP(), INTERVAL -24 HOUR);
```

## Minor changes
- The tests now only run when changes are made to `/commons` or `/backend`. 
- Created a folder structure for dex project. I tried to stay on a dex.ts file however it was getting extremely unwieldy. Let me know if a single file or another file structure is better.
- For typescrypt support, i just developed it using strict setting, and then removed it so other files wouldn't break. Let me know if a specific TS rule is more appropriate (such as not having any's or unknowns).
- I just realized the endpoints naming convention is all over the place, let me know the ideal convention @italomlp @acharb 
- When making dex volume queries, i noticed that the results don't seem to be filtered by 24hrs. Fetching 24h, 48h or overall returns the same amount of data.

## Important points of attention
- I'm unsure how redis data is being cleaned out in production, currently our redis cache is just storing it there until it's manually removed. Do we have a CRON running or something on our production servers? If not, let me know and i'll clean the cache on a 24h basis on express.
- When fetching prices from stellar.expert the queries can take an unreasonable amount of time (after around 50 minutes i gave up testing locally). I'm not sure how long is reasonable for this cache to be generated but there are just too many assets to calculate, and each asset requires a request *2 ( for 24 an 48h). Right now with current bigquery results, i would need to make approximately 70.000 requests.
- When fetching prices for an asset, sometimes i get various connection errors, such as 
image
this can create inconsistent results. Not sure what we can do on that end however but i thought it was important to report it. Once an error like that happens we treat the price as 0 and essentially remove the asset from the sum, since we have no basis to calculate it.
- Prices are currently not being cached, since the results are inconsistent, i thought i'd be better to do that after we fix that. As soon as we reach a stable point i can push a new commit storing the price data.
- Also on fetching prices, should we ignore the "native" assets from the sum?
- Currently we are unable to test the DEX volumes endpoint, it would take potentially hours and make dozens of thousands of requests for stellar expert per run.

### CODE REVIEW

File: backend/src/v2v3/dex/data.ts [hunk 1]
```diff
@@ -0,0 +1,182 @@
...
+}
+
+export async function getTradeData() {
+  const tradesBase = (filter = "") => `
+    SELECT count(*) as data 
+    FROM ${bigQueryEndpointBase}.history_trades
+    ${filter};
+  `;
+  const tradesQuery24h = tradesBase(`
+    WHERE ledger_closed_at >= TIMESTAMP_ADD(CURRENT_TIMESTAMP(), INTERVAL -24 HOUR)
```
acharb, line 56: for these trades let's also filter out the AMM trades, adding the `AND (selling_liquidity_pool_id IS NULL OR selling_liquidity_pool_id = '') ` clause

backend/src/routes.ts
```diff
@@ -48,6 +49,13 @@ app.get(
   lumensV2V3.v3CirculatingSupplyHandler,
 );
 
+app.get("/api/v2/dex/24h-payments", dex.get24hPaymentsData);
+app.get("/api/v2/dex/24h-trades", dex.getDexTrades24hData);
+app.get("/api/v2/dex/unique-assets", dex.getUniqueAssets);
+app.get("/api/v2/dex/volume", dex.getVolume);
+app.get("/api/v2/dex/active-accounts", dex.getActiveAccounts);
+app.get("/api/v2/dex/all", dex.getAll);
```
acharb, line 57: do we need this /all endpoint?

backend/src/routes.ts
```diff
@@ -48,6 +49,13 @@ app.get(
   lumensV2V3.v3CirculatingSupplyHandler,
 );
 
+app.get("/api/v2/dex/24h-payments", dex.get24hPaymentsData);
```
acharb, line 52: we don't need to use `v2` for these new ones. Let's keep em as `/api/dex/...`

the v2/v3 endpoints were added because we made some changes to endpoints that were being used by 3rd parties (ideally we'd like to clean up / deprecate the older versions, but need to look into if they're being used still first). Since these are new endpoints we can consider them independent of the other ones so don't need versions

backend/src/v2v3/dex/data.ts
```diff
@@ -0,0 +1,182 @@
+import { bqClient } from "../../bigQuery";
```
acharb, line 1: similarly let's move this logic out of `v2v3`, so would be at `backend/src/dex`

backend/src/v2v3/dex/data.ts
```diff
@@ -0,0 +1,182 @@
...
+    const [result] = await job.getQueryResults();
+    return result;
+  } catch (error: unknown) {
+    const message = getErrorMessage(error);
+    console.error("QUERY ERROR: ", message);
+    return null;
+  }
+}
+
+async function fetchMemoizedData(key: string, query: string) {
```
acharb, line 20: let's just call this cached? even though memoized is technically also correct we're using `cached` throughout the codebase
```suggestion
async function fetchCached(key: string, query: string) {
```

acharb (commented):
looking great! left some comments

acharb:
> I'm unsure how redis data is being cleaned out in production

right now we're not clearing redis but overwriting based on a schedule. (see here, this triggers the lumens cache updates every 10 mins)

I think for these 24HR redis updates, we should clear the redis on a schedule though. We could add a github action schedule workflow? That way we can say that the data is updated at everyday at a certain time, as opposed to when the servers re-deploy. I'd have to think more about best way of doing that in the code 🤔 , unless you have ideas how to do

acharb:
> When making dex volume queries, i noticed that the results don't seem to be filtered by 24hrs. Fetching 24h, 48h or overall returns the same amount of data.

let's remove the `selling_liquidity_pool_id = ''` clause, it's what's messing up the query (tbh not sure why, BQ bug?). But I checked and there aren't any entries w/ `selling_liquidity_pool_id = ''`.

So the dex volume queries should now look like:
(also let's use the `batch_run_date` for time partitioning)
```
...
SELECT SUM(buying_amount) AS summed_amount, CONCAT(buying_asset_code, ":", buying_asset_issuer) AS asset_name,
FROM `crypto-stellar.crypto_stellar_2.history_trades`
    WHERE selling_liquidity_pool_id IS NULL AND batch_run_date >= DATE_SUB(CURRENT_DATE(), INTERVAL 1 DAY) 

...

File: backend/src/v2v3/dex/data.ts (hunk 1 above)
Comment (Selhar) [reply backend/src/v2v3/dex/data.ts:56]: Should we remove the pool_id = '' from this one as well?

**Comment by Selhar** [comment]
> I think for these 24HR redis updates, we should clear the redis on a schedule though. We could add a github action schedule workflow? That way we can say that the data is updated at everyday at a certain time, as opposed to when the servers re-deploy. I'd have to think more about best way of doing that in the code thinking , unless you have ideas how to do

I think the best solution depends on how this backend will be hosted really. For a 24h cycle i don't think express is the best option for us because it can be unreliable and difficult to track in case of failures such as a server restart/deploy or errors. Using a CRON would allow us to make consistent calls directly to redis to clear cache on demand and have logs in case of a failure.

But that really depends on how much access we have to the redis service and how it's hosted. If we're running a docker compose somewhere, i might be able to create a service that mimicks a CRON job, but it won't be as good as the real thing. As for the github action i don't think that's ideal unless we don't have any other means to access the server (i've been in a similar situation where github actions triggered a cache invalidation and it was very unreliable, but it can work).

By the way, regarding the errors on stellar.prices, is there something i can do on my end to solve the throttling errors? Or is that something that won't happen in production? The new volume calls greatly decreased the runtime but it's still quite a long time for a cache refresh (since we still have to make the big overall query).

**Comment by acharb** [comment]
Good points, I forgot we have a couple cron jobs already set up using kubernetes. That's probably the best route here as well. I created a ticket to track.

Our kubernetes repo is an internal one I believe, so I think this is something an SDF person will have to do:
https://github.com/stellar/dashboard/issues/277

**Comment by acharb** [comment]
> By the way, regarding the errors on stellar.prices, is there something i can do on my end to solve the throttling errors?

Once we add the production API key it may resolve. We have other projects using the API key that are making similar calls. We'll have to add that internally.

Created a ticket to track: https://github.com/stellar/dashboard/issues/278

**Comment by sydneynotthecity** [comment]
One item that you may want to consider for DEX trading volume is to narrow down the list of assets. You could create a heuristic where the asset must have had N number of trades executed in 24 hours or a particular number of trustlines established before it should be part of the pricing pipeline. There are a lot of junk assets on the network and you don't want to unnecessarily assign them value. We're having discussions with `/path-payments` about similar ideas to narrow down the list of assets to make these calls more efficient.

File: backend/src/v2v3/dex/data.ts [hunk 6]
```diff
@@ -0,0 +1,182 @@
...
+
+export async function getPaymentsData() {
+  const query = `
+    SELECT count(*) as data 
+    FROM ${bigQueryEndpointBase}.enriched_history_operations 
+    WHERE type in (1, 2, 13) 
+    AND closed_at >= TIMESTAMP_ADD(CURRENT_TIMESTAMP(), INTERVAL -24 HOUR) 
+    AND successful is true
+  `;
+  const payments = await fetchMemoizedData("dex-payments24h", query);
```
sydneynotthecity, line 44: Does the dashboard have a similar label for `dex-payments-24h`? Technically non path-payment operations of type 1 do not go through the DEX to execute.

Selhar:
> There are a lot of junk assets on the network and you don't want to unnecessarily assign them value.

I'm making optimizations to the prices caching right now to make it easier when redis is cleared. If there are any thresholds i can use to reduce the amount of queries to stellar.expert, it'd be greatly appreaciated. Even with the improved dex volume queries it takes >30 minutes to make all requests and even then more than 1/3 of them fail. I don't really have a parameter for what is an acceptable threshold for volumes so i think it's best to leave for you folks.

backend/src/v2v3/dex/data.ts
```diff
@@ -0,0 +1,182 @@
...
+    SELECT count(*) as data 
+    FROM ${bigQueryEndpointBase}.history_trades
+    ${filter};
+  `;
+  const tradesQuery24h = tradesBase(`
+    WHERE ledger_closed_at >= TIMESTAMP_ADD(CURRENT_TIMESTAMP(), INTERVAL -24 HOUR) 
+    AND batch_run_date >= DATE_SUB(CURRENT_DATE(), INTERVAL 2 DAY)
+  `);
+  const tradesQuery48h = tradesBase(`
+    WHERE ledger_closed_at >= TIMESTAMP_ADD(CURRENT_TIMESTAMP(), INTERVAL -48 HOUR)
```
sydneynotthecity, line 60: This could be a nit, but the comparison of past 24 hours to past 48 hours of trades is a little weird to me. I would have expected the comparison to be past 24 hours vs past 48-past 24 hours so that we're compared the same interval of trades. 

With the current definition, we should expect a minimum of 100% difference if we assume constant trading activity over time. @acharb do you know if this was how the old dashboard defined it?

sydneynotthecity (commented):
Had a couple questions regarding metrics definition. Alec had already pointed out some of the questions for filter logic using the `selling_liquidity_pool_id`.

backend/src/v2v3/dex/data.ts
Selhar: I'm currently using
```
AND batch_run_date <= DATE_SUB(CURRENT_DATE(), INTERVAL 1 DAY) 
AND batch_run_date >= DATE_SUB(CURRENT_DATE(), INTERVAL 2 DAY) 
```
for 48h, and 
```
AND batch_run_date >= DATE_SUB(CURRENT_DATE(), INTERVAL 1 DAY)
```
for 24hrs for both trades and volume queries, let me know if it's ideal

acharb: > @acharb do you know if this was how the old dashboard defined it?

this is the first time we're showing Dex data on the dashboard.

Ya  I think this equation needs some adjusting, to compare the last 24h with the previous 24h:

```
const difference = results24h[0]?.data - results48h[0]?.data;
const output = (difference / results24h[0]?.data) * 100;
```
should instead be:
```
const previous24h =  results48h[0]?.data - results24h[0]?.data;
const output = (previous24h - results24h[0]?.data) / results24h[0]?.data * 100;
```

backend/src/v2v3/dex/data.ts
acharb: oh ya  I guess this technically isn't "dex" data, it should just be grabbing the last 24h payments.

@Selhar we could probably change this label to just `payments24h`?

@sydneynotthecity (not sure if you have access to these designs), but these are the metrics were looking to grab:

Screen Shot 2022-05-26 at 11 19 01 AM

backend/src/v2v3/dex/data.ts
Selhar: Updated the equation on trades and dex data. I'll finish up testing and put up the commits in a few mins.

Selhar: Hey, using the updated equation i'm getting some really funky results. Let's say previous day = 42383, today = 75055, using this equation the result is 43%, but don't we want the percentage of increase/decrease? 

Using this as a calculator the results are very different: https://www.calculator.net/percent-calculator.html?c3par1=42383&c3par2=75055&ctype=3&x=86&y=23#pctdifference

Selhar:
Changes committed. Some important attention points though.

- Added a cache for the volume calculation, we have a 24h and 48h cache. Once the 48h cache is cleared, the 24h is moved to the 48h slot and the 24hrs cache is renewed. Let me know if this logic is sound, i made it thinking of a CRON that will clean the 48hrs cache every day.
- Added a clause to the `getOrThrow` function to allow it not to throw on some edge cases. I thought of changing the name or creating another function but it seemed unnecessary.
- We no longer store the volume query result on redis, it costs a lot of memory and that data was only used to calculate the volume, now we only store the calculated volume itself. Let me know if the query result is necessary in some context.
- Even limiting the amount of stellar.expert queries to 300 (100 first assets for 24h, 48 and overall) the request and first cache takes 3 minutes to be created. 24hrs of assets results in around 6000 items so a complete cache would possibly take hours to be created, this makes the volumes endpoint impossible to test meaningfully. I've removed the limit to 100 first results so it's technically ready for deploy, but we should keep an eye on how long this cache takes to be generated and how much cost it creates on the stellar.expert servers.
image
- Updated equation for increase from the previous day, results are strange so i'll update them once we have an agreement on what they should represent.

> Our kubernetes repo is an internal one I believe, so I think this is something an SDF person will have to do:

Here are the keys you'll need to clear on kubernetes on a 24h basis. Ideally we should not deploy before this is in place, otherwise we'll always get the cached values.
`dex-volume-sum-48h`
`dex-volume-sum-overall`
`dex-payments24h`
`dex-trades24h`
`dex-trades48h`
`dex-trades-overall`
`dex-uniqueAssets`
`dex-activeAccounts`

backend/src/v2v3/dex/data.ts
sydneynotthecity: Excellent, just making sure you're representing just `payments` instead of `dex-payments`. Looks good!

backend/src/v2v3/dex/data.ts
sydneynotthecity: I think that's because on the denominator you should be using the previous 24 hr result, not current day. We should also flip numerator to find the difference between current day - previous 24hr:
```
(results24h[0]?.data  - previous24h) / previous24h * 100
```

acharb: oops ya Syd's equation looks right, thanks for checking

backend/src/v2v3/dex/data.ts
Selhar: Fixed the label and looking into the updated query (hadn't seen the comment before). Pushing new commits shortly.

Selhar:
Hey, my tests aren't running, they usually fail whenever i run something that requests data from bigquery. Maybe we're missing the new environment variable on github actions @acharb, BQ_PROJECT_ID more specifically. (sorry, left the comment on the issue instead of the PR :facepalm: )

backend/src/v2v3/dex/data.ts
Selhar: Hey, the results are still very weird, i'll give an example scenario to make it easier to talk about the equation since i'm terrible at math.
Let's say current day is 16, and the previous is 15. 
Day 15 results: 20
Day 16 results: 10
Expected result: 50% decrease
From what i understand the formula is the same, we only need to know if day 16 is smaller than day 15 or not to give a positive or negative result. So our code could be as such:

```
const increase =  Math.abs(sum48h - sum24h)
const difference = increase / sum48h * 100
const change = sum24h > sum48h ? difference : -Math.abs(difference)
```
Applying the previous scenario to this formula, the results would be:
```
const increase =  Math.abs(20 - 10)
const difference = 10 / 20 * 100
const change = 10 > 20 ? 50 : -50
```
This would output `-50`

that's correct right? With the following data i get +77% change.
image

acharb: this equation should be right: `(results24h - previous24h) / previous24h * 100`
(eg. from your example results24 = 10, previous24h = 20, gives -50%)

The dex-trades values don't look right. dex-trades48h should be greater than dex-trades24h, since it's the last 48 hours of trades vs the last 24 hours.

You can think of it this way (using the redis labels):
```
previous24h = dex-trades48h - dex-trades24h
results24h = dex-trades24h

change = (results24h - previous24h) / previous24h * 100
```

Selhar: Wait but our queries are not cumulative. The 48hrs query fetches from 2 days ago excluding the last 24hrs. Here's an example of filter used:
```
    AND batch_run_date <= DATE_SUB(CURRENT_DATE(), INTERVAL 1 DAY) 
    AND batch_run_date >= DATE_SUB(CURRENT_DATE(), INTERVAL 2 DAY) 
```

acharb: ah sorry, I thought it was cumulative. So instead it would be:
```
previous24h = dex-trades48h
```
The % diff equation should be the same though

acharb: are the numbers weird with that equation? I think let's use it for now, if we find the % diffs are weird we can take a look at it and adjust later

Selhar: Don't we have to check for negative values though? In this example, if the 48hrs query is larger than the 24hrs query, it would break. If we apply this equation in a decrease scenario, won't it break?

Using this equation
```
```

Selhar: That makes sense! I'm comitting the changes, here are the results on a volume query just to make sure we're on the same page:

48h: `75055.36712705049`
24h: `42382.75063436342`
change: `-77.08941964280275`

is that correct?

sydneynotthecity: @Selhar that looks good to me! The `change` is what I would expect it to be given those values.

Selhar:
Hey, just a summary of the last things pending on this task:
- "native" assets are not included in the volume sum, should they be ignored (price = 0) or is there an asset we can use to calculate their volume?
- Will we add tests that make queries to bigquery? If not i can remove the ones i added in this task. Otherwise i think we're missing the environment variables.

Also if there are any changes to the code you folks would like to make before we merge, just let me know. In the meantime i'll be working on the ledgers task.

acharb:
@Selhar let's remove the dex volume code (ie the sum_volume and get_volume_data methods). See slack for more info: https://stellarfoundation.slack.com/archives/C023RFQKL85/p1654126069606199
tldr we'd like to do it not using stellar.expert, which we can't do atm

Selhar:
Hey just as an update, i'm finishing up the ledgers script and will close up the final changes we need for dex after that. Since it's a fairly complex script i'd rather get it done asap. Should be done later today.

Selhar:
- Removed the calls to the volume endpoint but left the function. In the future we should be able to use it, only changing the datasource from stellar.expert to the returned data itself. Let me know if it'd be better to remove the function entirely.
- Added mocked redis data for tests.

acharb (approved):
nice lgtm! had 1 request to add a comment

backend/src/dex/data.ts
```diff
@@ -0,0 +1,202 @@
...
+        });
+    },
+    0,
+  );
+
+  await redisClient.set(name, JSON.stringify(output));
+  return output;
+}
+
+export async function getVolumeData() {
```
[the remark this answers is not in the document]
Selhar: Done!

PR #282, opened 2022-06-08
SDFD-13_serve_1h_24h_ledger_data

Selhar:
Hey, this was a bit tricky. I tried various forms to implement this without disturbing much of the original code, this PR is the best attempt i've made with minimal changes to the original algorithm. Let me know if any of it is not ideal or otherwise inappropriate @acharb 
Some of the important changes:
- Changed the query to define our initial pointer from `select * ...` to `select id ...`, since we're only using ID and this was having a large impact on free account queries threshold. Sizes before/after:
image
image
- For the intervals, i created a data structure with values of `hour`, `day`, `month`. Various parts of the code needs different kinds of data depending on the interval we need, but if we're able to access all of them with the same key, an enum allows us to very easily make the code dynamic without much repetition or reestructuring of the original logic. Let me know if any part seems difficult to understand or counterintuitive.
- I removed the ledger fetch loop from `while(true)` to a recursive iteration, because a `while(true)` loop blocks the [event loop](https://nodejs.dev/learn/the-nodejs-event-loop), causing resources to hang until the operation completes. Creating a new scope allows the node engine to manage resources more easily. Side effects of a blocked event loop on the backend would be dynamic requests that don't answer until we finish a `catchup()` call.
- Maintained all original calls to 30 day ledger as is, all signatures and endpoints remain the same.
- Day/hour endpoints are `/api/ledgers/day/public` and `/api/ledgers/hour/public`
- Day endpoint returns data on an interval of 1h, therefore 24 datapoints maximum.
- Hour endpoints return on a minute to minute interval, therefore 60 datapoints.
- day interval sample:
```
[
    {
        "date": "06-07 5H",
        "transaction_count": 58341,
        "operation_count": 106878
    },
    {
        "date": "06-07 6H",
        "transaction_count": 224879,
        "operation_count": 477175
    },
]
```

- hour interval sample:
```
[
    {
        "date": "06-08 4H:45M",
        "transaction_count": 3477,
        "operation_count": 5594
    },
    {
        "date": "06-08 4H:46M",
        "transaction_count": 4797,
        "operation_count": 7660
    },
]
```
- Regarding tests, since the functionality itself hasn't changed, only static values depending on the interval, i couldn't see anything new worth testing, our current tests already cover all scenarios i can think of. Let me know if you miss any tests.

### DIFF
--- a/backend/src/ledgers.ts
+++ b/backend/src/ledgers.ts
@@ -31,6 +31,14 @@ const REDIS_PAGING_TOKEN_KEY = {
   month: "paging_token",
 };
 
+const HORIZON_LIMITS = {
+  hour: 10,
+  day: 100,
+  month: 200,
+};
+
+const intervalTypes = [INTERVALS.day, INTERVALS.hour, INTERVALS.month];
+
 const BIGQUERY_DATES = {
   hour: getBqDate(INTERVALS.hour),
   day: getBqDate(INTERVALS.day),
@@ -54,28 +62,22 @@ function getBqDate(interval: INTERVALS) {
 
 interface LedgerStat {
   date: string;
-  /* eslint-disable camelcase */
   transaction_count: number;
   operation_count: number;
-  /* eslint-enable camelcase */
 }
 
 // TODO - import Horizon type once https://github.com/stellar/js-stellar-sdk/issues/731 resolved
 export type LedgerRecord = {
-  /* eslint-disable camelcase */
   closed_at: string;
   paging_token: string;
   sequence: number;
   successful_transaction_count: number;
   failed_transaction_count: number;
   operation_count: number;
-  /* eslint-enable camelcase */
 };
 const CURSOR_NOW = "now";
 
 export async function handler(_: any, res: Response, next: NextFunction) {
-  // - attempt to fetch cached redis data
-  // - return it
   try {
     const cachedData = await getOrThrow(redisClient, REDIS_LEDGER_KEYS.month);
     const ledgers: LedgerStat[] = JSON.parse(cachedData);
@@ -85,9 +87,27 @@ export async function handler(_: any, res: Response, next: NextFunction) {
   }
 }
 
-export async function updateLedgers() {
-  const intervalTypes = [INTERVALS.day, INTERVALS.hour, INTERVALS.month];
+export async function handler_day(_: any, res: Response, next: NextFunction) {
+  try {
+    const cachedData = await getOrThrow(redisClient, REDIS_LEDGER_KEYS.day);
+    const ledgers: LedgerStat[] = JSON.parse(cachedData);
+    res.json(ledgers);
+  } catch (e) {
+    next(e);
+  }
+}
 
+export async function handler_hour(_: any, res: Response, next: NextFunction) {
+  try {
+    const cachedData = await getOrThrow(redisClient, REDIS_LEDGER_KEYS.hour);
+    const ledgers: LedgerStat[] = JSON.parse(cachedData);
+    res.json(ledgers);
+  } catch (e) {
+    next(e);
+  }
+}
+
+export async function updateLedgers() {
   for (const interval of intervalTypes) {
     const cachedData =
       (await redisClient.get(REDIS_LEDGER_KEYS[interval])) || "[]";
@@ -96,7 +116,6 @@ export async function updateLedgers() {
 
     if (cachedLedgers.length < LEDGER_ITEM_LIMIT[interval]) {
       try {
-        console.log("DEBUG - INTERVAL: ", interval);
         const query = getBqQueryByDate(BIGQUERY_DATES[interval]);
         const [job] = await bqClient.createQueryJob(query);
         console.log("running bq query:", query);
@@ -118,6 +137,7 @@ export async function updateLedgers() {
       pagingToken,
       REDIS_PAGING_TOKEN_KEY[interval],
       0,
+      interval,
     );
   }
 
@@ -141,51 +161,45 @@ export async function updateLedgers() {
 
 export async function catchup(
   ledgersKey: string,
-  pagingTokenStart: string,
+  pagingToken: string,
   pagingTokenKey: string,
   limit: number,
+  interval: INTERVALS = INTERVALS.month,
+  total: number = 0,
 ) {
-  // using the starting point defined previously, call for horizon records
-  // until the records end or we reach a record limit
-  // set the pointer towards the latest ledger we got
-  // update the cache with received ledgers
   const horizon = new stellarSdk.Server("https://horizon.stellar.org");
+  const resp = await horizon
+    .ledgers()
+    .cursor(pagingToken)
+    .limit(HORIZON_LIMITS[interval])
+    .call();
   let ledgers: LedgerRecord[] = [];
-  let total = 0;
-  let pagingToken = pagingTokenStart;
 
-  // eslint-disable-next-line no-constant-condition
-  while (true) {
-    // eslint-disable-next-line no-await-in-loop
-    const resp = await horizon.ledgers().cursor(pagingToken).limit(200).call();
-    ledgers = resp.records;
-    total += resp.records.length;
-    if (ledgers.length === 0 || (limit && total > limit)) {
-      break;
-    }
+  ledgers = resp.records;
+  total += resp.records.length;
 
-    pagingToken = ledgers[ledgers.length - 1].paging_token;
-    // eslint-disable-next-line no-await-in-loop
-    await updateCache(ledgers, ledgersKey, pagingTokenKey);
+  if (ledgers.length === 0 || (limit && total > limit)) {
+    return;
   }
+  pagingToken = ledgers[ledgers.length - 1].paging_token;
+  await updateCache(ledgers, ledgersKey, pagingTokenKey, interval);
+
+  await catchup(
+    ledgersKey,
+    pagingToken,
+    pagingTokenKey,
+    limit,
+    interval,
+    total,
+  );
 }
 
 export async function updateCache(
   ledgers: LedgerRecord[],
   ledgersKey: string,
   pagingTokenKey: string,
+  interval = INTERVALS.month,
 ) {
-  // - check for amount of ledgers
-  // - fetch cached ledgers from redis
-  // - iterate each of the received ledgers (not the ones already cached)
-  // - attempt to find a stored value for that day by it's key (the date formatted)
-  // - If you do not find that day, push new data into the array
-  // - If you do find a result, switch the new ledger in place of the old one. Sum the operation count for the day.
-  // - set the pointer towards the latest stored ledger
-  // - Sort the ledgers by most recent
-  // - store the 30 most recent ledgers
-  // - set the paging token redis variable
-  // - log the ledge update
   if (!ledgers.length) {
     console.log("no ledgers to update");
     return;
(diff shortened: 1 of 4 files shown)


Selhar:
As a suggestion, i think the stellar SDK could answer for time intervals ("give me all ledgers between these 2 dates on an interval of X minutes"), this would greatly reduce the cost of these operations since a dashboard is probably the first thing a third party might want to build. On an empty cache, trying to fill the monthly ledger is an enormous and extremely expensive task, since the stellar query limit is 200, and 200 results represents only a minute or so and we have to fill an entire month worth of data.

acharb (commented):
some initial small comments.

Also, when I run it on my local, it stops updating w/ new ledgers after it's caught up. Is it working for you?

```
$ UPDATE_DATA=true DEV=true ts-node ./src/app.ts
starting ledgers cache update
connected to redis
ledgers updated to: 2022-06-09T00:56:56Z
ledgers updated to: 2022-06-09T00:57:02Z
ledgers updated to: 2022-06-09T00:57:09Z

... ^ stops w/ no more updates
```

Selhar:
> Also, when I run it on my local, it stops updating w/ new ledgers after it's caught up. Is it working for you?

Yeah, it doesn't work here but it never has. When i started this task i tested it out in the `dashboard-v2` branch and it didn't work. I assumed it was due to local environment. If you go to the main branch, can you get constant updates? The logs below are from the `dashboard-v2` branch, it has been running for around 30min.

image

backend/src/ledgers.ts
```diff
@@ -5,37 +5,101 @@ import { Response, NextFunction } from "express";
...
   failed_transaction_count: number;
   operation_count: number;
-  /* eslint-enable camelcase */
 };
-
-const REDIS_LEDGER_KEY = "ledgers";
-const REDIS_PAGING_TOKEN_KEY = "paging_token";
 const CURSOR_NOW = "now";
 
 export async function handler(_: any, res: Response, next: NextFunction) {
```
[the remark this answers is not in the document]
Selhar: should the endpoint also change for /ledgers/month/public? or do we leave it as ledgers/public?

acharb:
ah I found out why, there's a bug with v10.1.0, which is resolved in 10.1.1. Updated on dashboard-v2: https://github.com/stellar/dashboard/pull/285

backend/src/ledgers.ts
```diff
@@ -176,3 +263,13 @@ function formatDate(s: string): string {
   const day = `0${d.getUTCDate()}`;
   return `${month.slice(-2)}-${day.slice(-2)}`;
 }
+
+const getLedgerKey = {
+  hour: (closed_at: string) =>
```
acharb, line 268: for the hours can we use just the minute as the key? this will make it easier for both the frontend to use and for debugging
eg. it would be:

```
[
  {
    "date": "57M",
    "transaction_count": 703,
    "operation_count": 1721
  },
  {
    "date": "56M",
    "transaction_count": 3833,
    "operation_count": 9317
  },
  {
    "date": "55M",
    "transaction_count": 3264,
    "operation_count": 8717
  },
```

backend/src/ledgers.ts
```diff
@@ -176,3 +263,13 @@ function formatDate(s: string): string {
...
+    `${formatDate(closed_at)} ${new Date(closed_at).getUTCHours()}H:${new Date(
+      closed_at,
+    ).getUTCMinutes()}M`,
+  day: (closed_at: string) =>
+    `${formatDate(closed_at)} ${new Date(closed_at).getUTCHours()}H`,
```
acharb, line 273: similarly let's get rid of the date for the day endpoint, so would be like:
```
  {
    "date": "0H",
    "transaction_count": 35109,
    "operation_count": 69483
  },
  {
    "date": "1H",
    "transaction_count": 269786,
    "operation_count": 471790
  },
  {
    "date": "2H",
    "transaction_count": 278009,
    "operation_count": 446415
```

backend/src/ledgers.ts
```diff
@@ -75,42 +148,57 @@ export async function updateLedgers() {
...
+    interval,
+    total,
+  );
 }
 
 export async function updateCache(
   ledgers: LedgerRecord[],
   ledgersKey: string,
   pagingTokenKey: string,
+  interval = INTERVALS.month,
```
acharb, line 201: This is causing an error looks like when streaming new ledgers. I suggest we remove the default values for interval, to make sure the right one is getting passed in
```suggestion
  interval: INTERVAL,
```

backend/src/ledgers.ts
```diff
@@ -75,42 +148,57 @@ export async function updateLedgers() {
     .limit(200)
     .stream({
       onmessage: async (ledger: LedgerRecord) => {
-        await updateCache([ledger], REDIS_LEDGER_KEY, REDIS_PAGING_TOKEN_KEY);
+        for (const interval of intervalTypes) {
+          await updateCache(
+            [ledger],
+            REDIS_LEDGER_KEYS[interval],
+            REDIS_PAGING_TOKEN_KEY[interval],
+          );
```
acharb, line 156: need to pass in `interval` here

Selhar:
> the dates in the endpoint should go from most recent to least recent.

Oh, i completely forgot to mention that, my bad. I was talking with @italomlp yesterday about that, and one solution we could use is using a datetime with zeroes in place of unused parts of the date. Such as:
Month: MM-DD 00:00:00 (since we sort by day)
Day: MM-DD HH:00:00 (since we sort by hour)
Hour: MM-DD HH:MM:00 (since we sort by minute)

That way the date sorting script wouldn't have to change since it's a valid subtraction still. The reason i was holding on that is because i didn't want to mess with the original implementation of "month". What do you think of this approach?

acharb:
> The reason i was holding on that is because i didn't want to mess with the original implementation of "month". What do you think of this approach?

I wouldn't be too tied to the original "month" implementation. Since we're adding "hour" and "day" now we can go with a solution that makes sense for all of them.

I'm a little hesitant of using a date time as the keys for the data objects. Since these will always be changing I think it'll make it more difficult to handle? (for both updating when getting a new ledger and also for the frontend to handle). But up to you and @italomlp , if you guys think this way is simpler/easier then sgtm!

Selhar:
> think it'll make it more difficult to handle

Oh, we'll only partially use datetime. We'll manually add 0's in place of the unused data pieces, that way they stay consistent with what we have today. I'll give it a try and send in the implementation.

Selhar:
https://twitter.com/teamspeak/status/1534890773002452992

Selhar:
Hey, pushing a small commit with sequence alongside ledger data so @italomlp can use it on a frontend tooltip @acharb, more info here: https://cheesecake.slack.com/archives/C023RFQKL85/p1654802113867269

Selhar:
I think this solves all issues, could you take another look @acharb? Only thing missing is averages data, since @italomlp doesn't have access to closed_at and other metadata, i'll provide averages for transactions, operations and closed times on this same payload.

backend/src/ledgers.ts
```diff
@@ -5,37 +5,102 @@ import { Response, NextFunction } from "express";
...
+const LEDGER_ITEM_LIMIT = {
+  hour: 60,
+  day: 24,
+  month: 30,
+};
+
+const REDIS_LEDGER_KEYS = {
+  hour: "ledgers_hour",
+  day: "ledgers_day",
+  month: "ledgers",
```
acharb, line 25: let's make this month to to be consistent
```suggestion
  month: "ledgers_month",
```

backend/src/ledgers.ts
```diff
@@ -5,37 +5,102 @@ import { Response, NextFunction } from "express";
...
+};
+
+const REDIS_PAGING_TOKEN_KEY = {
+  hour: "paging_token_hour",
+  day: "paging_token_day",
+  month: "paging_token_month",
+};
+
+const HORIZON_LIMITS = {
```
acharb, line 34: is there a reason we need different horizon limits for the different intervals? We can't use `200` for each?

backend/src/ledgers.ts
```diff
@@ -145,10 +237,9 @@ export async function updateCache(
   });
   cachedStats.sort(dateSorter);
 
-  // only store latest 30 days
   await redisClient.set(
     ledgersKey,
-    JSON.stringify(cachedStats.slice(0, LEDGER_DAY_COUNT)),
+    JSON.stringify(cachedStats.slice(0, LEDGER_ITEM_LIMIT[interval])),
   );
   await redisClient.set(pagingTokenKey, pagingToken);
```
acharb, line 245: we can remove some of the code w/in `dateSorter` since the object keys are using timestamps now. Some of that logic was there because of edge cases of comparing dates w/o the year when they were in the format like:
```
[
  {
    "date": "06-13",
    "transaction_count": 4894311,
    "operation_count": 7576618
  },
  {
    "date": "06-12",
    "transaction_count": 5055858,
    "operation_count": 10093060
  },
  {
    "date": "06-11",
    "transaction_count": 5113215,
    "operation_count": 9201338
  },
...
```

The dateSorter function can just be this now, thanks to the timestamp key idea you guys had 😄 :
```
function dateSorter(a: LedgerStat, b: LedgerStat) {
  return new Date(b.date).getTime() - new Date(a.date).getTime();
}

```

acharb (commented):
nice! just had some final small comments and 1 question. I think after these changes we'll be able to merge 👍

backend/src/ledgers.ts
Selhar: Oh, for time intervals. Since we always get the last result from each query, 200 is too much for an hour or a day and determining the time interval in a batch of 200 would be more complex than seems worth it. Considering that if we're only getting the last hour, an interval of 10 isn't much of a downside, it's still going to be really fast to get through it. Same for day.

backend/src/ledgers.ts
Selhar: True, good catch!

Selhar:
Hey, pushing a WIP since this seemingly small change required a huge overhaul of our current system. Some things are still missing (tests and redis cache cleanup/optimization, might be missing some typescript types as well) but overall feel free to start the review process. As soon as they are finished i'll post more detailed info as to why this was required. I separated the commits as much as possible so the review won't be as painful. Changing the folder structure was mostly to organize my thoughts since this required massive amounts of data wrangling and one file was getting unwieldy.

@italomlp here's a sample of the output, let me know if anything needs adjustement. Feel free to get my branch and start integration with the frontend. Output won't change from now on.
image

acharb:
hey @Selhar I'm trying to run this and getting the error:
```
/Users/aleccharbonneau/Workspace/sdf/dashboard/backend/src/ledgers/data.ts:5
import { bqClient, BQHistoryLedger, getBqQueryByDate } from "../bigQuery";
                                                         ^
TypeError: Cannot read properties of undefined (reading 'transaction_count')
    at /Users/aleccharbonneau/Workspace/sdf/dashboard/backend/src/ledgers/data.ts:213:37
    at Array.forEach (<anonymous>)
    at /Users/aleccharbonneau/Workspace/sdf/dashboard/backend/src/ledgers/data.ts:175:11
    at Generator.next (<anonymous>)
    at fulfilled (/Users/aleccharbonneau/Workspace/sdf/dashboard/backend/src/ledgers/data.ts:5:58)
```
is this running for you?

acharb:
also can we merge in the work you did before adding these average changes? I think that code was good to go. It also makes seeing the average changes you made more clearly, and keeps the PRs a more manageable size 😄

Selhar:
> is this running for you?

Yeah, maybe you have a stale cache? Try flushing current cache and then creating a new one. The tests we have on github actions also ran without issue.



> also can we merge in the work you did before adding these average changes?

Sure! I'll edit this PR and create a new one!

PR #287, opened 2022-06-16
SDFD-13_serve_ledger_averages

Selhar:
On the front end we need averages data based on the output of all ledges. Since we do not normally use all ledgers, only those of an interval (1min, 1h, 1 day) this required a considerable restructuring of our systems.

More context is in #282


Selhar:
After approval we need a new cache for the preview environment to work. Also, if you need to do further testing, hop on this branch instead @italomlp

backend/src/ledgers/data.ts
```diff
@@ -31,42 +33,31 @@ const REDIS_PAGING_TOKEN_KEY = {
...
+  month: "ledgers_month",
 };
 
-const intervalTypes = [INTERVALS.day, INTERVALS.hour, INTERVALS.month];
-
-const BIGQUERY_DATES = {
-  hour: getBqDate(INTERVALS.hour),
-  day: getBqDate(INTERVALS.day),
-  month: getBqDate(INTERVALS.month),
+export const REDIS_AVGS_LEDGER_KEYS = {
```
acharb, line 42: doesn't look like `REDIS_AVGS_LEDGER_KEYS` is used. Are they meant to be? Otherwise let's remove

acharb, line 42: doesn't look like `REDIS_AVGS_LEDGER_KEYS` is used. Are they meant to be? Otherwise let's remove

backend/__tests__/lumens.test.ts
```diff
@@ -199,15 +200,55 @@ describe("ledgers", () => {
       expect(JSON.parse(cachedLedgers as string)).toEqual([
         {
           date: "2022-1-12 00:00:00",
-          transaction_count: 80,
-          operation_count: 300,
-          sequence: 10003,
+          data: {
```
acharb, line 203: instead of storing the average data for each day, couldn't we just store the total number of ledgers and total duration from first ledger to last ledger for that day? (and same would apply for each minute in the `hour` data and each hour in the `day` data points) Then in `formatOutput` it could compute the averages of all the days then.

I may not be thinking of something though

acharb, line 203: I might be missing something, but why does each day here need to store that day's average?

couldn't we just store the total number of ledgers and total duration from first ledger to last ledger? Then in `formatOutput` it could compute the averages of all the days then.

acharb: so cachedLedgers would instead look something like:
```
{
  date: "2022-1-12 00:00:00",
  data: {
    transaction_count: 80,
    operation_count: 300,
    sequence: 10003,
    totalLedgers: 500,
    totalDuration: 3600
  },
},
{
  date: "2022-1-11 00:00:00",
  data: {
    transaction_count: 15,
    operation_count: 50,
    sequence: 10001,
        totalLedgers: 500,
    totalDuration: 3600
  },
},
```

acharb: so cachedLedgers would instead look something like:
```
{
  date: "2022-1-12 00:00:00",
  data: {
    transaction_count: 80,
    operation_count: 300,
    sequence: 10003,
    totalLedgers: 500,
    totalDuration: 3600
  },
},
{
  date: "2022-1-11 00:00:00",
  data: {
    transaction_count: 15,
    operation_count: 50,
    sequence: 10001,
        totalLedgers: 500,
    totalDuration: 3600
  },
},
```

Selhar: That's a great catch! I made a poc and the results seem consistent. I'll make the changes. This also will solve the issues with in-memory cache sizes.

backend/src/ledgers/data.ts
Selhar: Oh, completely forgot about this one. It was intended to store the processed values so we didn't have to permanently store thousands of timestamps on redis. I'll remove it.

backend/__tests__/lumens.test.ts
Selhar: Hey, actually the numbers are a bit inconsistent. 
Day ledgers returns
image
If i apply this logic:
image
I get this result
image

Full calculation: 5.9secs
First - last / size: 5.4secs

Same happens for last hour calculation:
Full calculation: 5.3secs
First - last / size: 5.7secs

Given that one result is higher and the other smaller, i'm not sure if my algorithm is incorrect or this loss of precision is expected given the simplification. Should we go ahead with this algorithm? Is this in the margin of error (0.5secs~)?

Selhar: Oh it's literally just that. The initial values had an incomplete cache and so were accurate, after you fill it up it the distance increases. Waiting on 30d cache to fill to see if it gets to an unreasonable amount.

Here's the code, just paste it at the start of `formatOutput()`
```
  const average_sums = sumAverages(cachedLedgers);

  const first = cachedLedgers[0].averages.closed_times_avg.sum[0]
  const last = cachedLedgers.at(-1).averages.closed_times_avg.sum.at(-1)
  console.log("DEBUG:", (first - last) / average_sums.closed_times_avg.size)
```

Selhar: Yeah the difference remains in the 0.5 range even for monthly ledgers, so i'll procceed with the implementation. We can improve the calculations later on if it becomes necessary.

Selhar: Hey, i was manually calculating the values and this doesn't really work. The issue is that if we take the initial and the final times we're assuming that each ledger closed at a uniform pace with no variation. @quietbits made the original script so she might be able to give some insight here.

To get an actual average we have to calculate every single one against the next, otherwise the values will never be accurate. They might sometimes be accurate if the network is stable and the results are pretty much even (which is why i got lucky earlier and the results were within the margin of error), but otherwise.. not so much. I got some absurd results like 170secs average, or 1.7secs. And they are correct...if you assume an even distribution of ledgers close time.

acharb: wanted to double check and it looks like it's working fine for me, lmk if there's something I'm missing though
see here https://github.com/stellar/dashboard/pull/288

acharb: ah woops I left out splitting failed and successful transactions. But the idea is the same

feel free to merge and build off of or just add to this PR

Selhar: Hey, it's still giving wildly inaccurate results, i think my lack of math savvy made things more convoluted than they had to be, but i don't think we can get ((last - first) / total) / 1000, here's an example, let's say we have the following timestamps:
1655725906133
1655725916133
1655725926133

there are 10 seconds between each of these timestamps, so 10 seconds is the result we want for our calculation.
If we apply this formula, we'll get the wrong results: 
```
let initial = 1655725906133
let final = 1655725926133
let output = ((final - initial) / 3)/1000
/// output = 6.666666666666667
```
because a unit of time for us is not a timestamp, but an interval between two timestamps. So our operation to get the correct result would be
```
let arraySum = [
[(1655725916133 - 1655725906133) / 1000, 1],
[(1655725926133 - 1655725916133) / 1000, 2]
]
let output = arraySum.reduce((prev, next) => (prev + next[0]) / next[1], 0)
//output = 10
```

Does that make sense? This is why we must store an array instead of the resulting operation, because we can only start the calculation once we have complete arrays, not partials.

Selhar: Hey, i sent a new commit with an optimized method of calculation, i manually checked the results and they are accurate for month, day and hour. This will consume a lot less memory since we're not storing the entire array and is still accurate, the formula is basically the one i sent in the comment above.

acharb: yup it should be `((last - first) / (total - 1)) / 1000`, we don't want to count the first ledger good catch

Selhar: i'm not sure i get it, we would still have an issue with averages because last - first doesn't represent the interval between ledgers. The -1 from total would work in a 3 number sample size but not in anything larger. For example:
```
let first = 1655725906133
let second = 1655725916133
let third = 1655725926133
let fourth = 1655725936133
```
```
let initial = first
let final = fourth
let output = ((final - initial) / 4-1)/1000
/// output = 7.499
```

The problem is not the total calculation, it's the individual unit of calculation that `((last - first) / (total - 1)) / 1000` represents.

backend/src/ledgers/utils.ts
```diff
@@ -0,0 +1,104 @@
...
+  const first = new Date(cachedLedgers.at(-1).data.start).getTime();
+  const last = new Date(cachedLedgers[0].data.end).getTime();
+  const avg_close_time = (last - first) / total_ledgers / 1000;
+  const avg_successful_tx_count = total_successful_tx / total_ledgers;
+  const avg_failed_tx_count = total_failed_tx / total_ledgers;
+  const avg_op_count = total_ops / total_ledgers;
+
+  const output = { data: [] };
+  cachedLedgers.forEach((ledger) => {
+    output.data.push({
```
acharb, line 91: just noticied this now. Any reason we can't just store the data in redis in the same data format as we're returning to the endpoint? Will make understanding and handling the data structure simpler I think if they're consistent

doesn't have to be as part of this PR, but lmk what you think

acharb (approved):
this is great lgtm!

I had one thought, non-blocking for this PR though

backend/src/ledgers/utils.ts
Selhar: Ye that would be ideal, but i'm not sure how to do it. Let's say we have 10 ledgers, each request retrieves 5, so at the end of the first request we would have a cache with the output format ready, something like:
```
{
  ...
  avg_successful_tx: 2.7
}
```

Now that we have that cache, when we move on to the second request, how can we calculate the remaining 5 items? Since we lost the original sum. Possibly we could store the final output + the partial sum, such as:
```
   date: ..,
   data: { ... },
   output: { ... },
```
but that would create a lot of processing overhead, since we have to re-calculate the output for every updateLedgers call. Let me know what you think, i don't have any more backend tasks for 2.1, so i'm pretty free to do any refactoring.

PR #288, opened 2022-06-17
use total sums to find the averages in formatOutput

acharb:


acharb:
@Selhar this is missing splitting the transaction count into successful and failed ones, but the idea should for the averages is the same. Feel free to close and use your PR or merge and build off this one

Selhar:
Yeah this was the implementation i had here, but it creates the problem where we're hiding averages. So if we have 200 transactions, we're assuming all of them took exactly the same time which is fine for 1h interval but the higher the intervals, the more wildly different the values will be from reality.

Selhar:
image

Final output sample. Feel free to merge after reviewing, this has already been rebased with #287.

acharb:
awesome thanks for fixing it 💯

Issue #273, opened 2022-05-24
Fee Module 


jcx120:
Build a "Fees" container component with the following elements:

1. Average Transaction Fee: 0.0002 XLM
   - [Reference example](https://messari.io/asset/stellar/chart/txn-fee-avg)
2. Graph of Average Transaction Fee over time (24hr, 30d)
   - May need to be derived from Hubble data and stored in time series
4. Base operation fee: 0.00001 XLM 
   - Src: static variable from code
5. Base reserve: 0.5 XLM
   - Src: static variable from code  
6. Fee Stats (Last 5 Ledgers) 
   - Metric | Value

Considerations:

- The "Fees" container component will need to be created
- Use [Stellar Design System](https://design-system.stellar.org/) (SDS) components as much as possible.
- Icons should also be used from the SDS (<Icon /> component). It's OK if they differ slightly.
- Backend data will vary depending on the data element
- Only on Mainnet (no Testnet).

raphaelmartins1993:
[jcx120](https://github.com/jcx120)
**Average Transaction Fee: 0.0002 XLM**

> Could you send us more information about the calc to this average?

**Graph of Average Transaction Fee over time (24hr, 30d)**

- May need to be derived from Hubble data and stored in time series

> Is it possible to bring us more context about this data source from Hubble? Which tables, etc.

3. Base operation fee: 0.00001 XLM
4. Base reserve: 0.5 XLM

> These two information are fixed?

italomlp:
Here are some screenshots of the possible final chart for this Fee module. Feedback is much appreciated, maybe for changing some colors, spacings, etc. LMK your thoughts @jcx120

<details>
<summary>Click to see screenshots:</summary>

PS: I'm using fake values, don't take them into account.

<img width="785" alt="Screen Shot 2022-06-09 at 15 04 21" src="https://user-images.githubusercontent.com/17295921/172918304-edc7322a-95ef-420c-a290-2a4363af5a6d.png">
<img width="781" alt="Screen Shot 2022-06-09 at 15 04 30" src="https://user-images.githubusercontent.com/17295921/172918315-3da46466-8004-4a93-b9ee-74f31cfc18d4.png">
<img width="766" alt="Screen Shot 2022-06-09 at 15 04 40" src="https://user-images.githubusercontent.com/17295921/172918340-24a169bc-b4ed-4a46-9c31-2ac28b40d417.png">
[...]

PR #292, opened 2022-06-27
SDFD-26: feat fee stats table

lccoronel:
1. Average Transaction Fee: 0.0002 XLM

2. Base reserve: 0.5 XLM

     * Src: static variable from code

3. Base operation fee: 0.00001 XLM

     * Src: static variable from code

Screen: 

Screen Shot 2022-07-19 at 10 59 33

4. Graph of Average Transaction Fee over time (24hr, 30d)

5. For Average Transaction Fee, we can use this sample query in Hubble to get the average over the last 30 days:

Screen:

Screen Shot 2022-07-19 at 10 59 25

6. Fee Stats (Last 5 Ledgers)

     * Metric | Value
     
Screen:

Screen Shot 2022-07-19 at 10 59 48


acharb:
are there mockups or a requirements list for adding these fee stats? just curious

frontend/src/components/AverageTransactionFee/index.tsx
```diff
@@ -0,0 +1,94 @@
...
+      titleCustom={timeRangeOptions}
+      noData={!feeStats.fees}
+    >
+      <div className="AverageTransactionFee__mainChart">
+        <div className="AverageTransactionFee__mainChart__container">
+          <VerticalBarChart
+            data={data[rangeInterval as "day" | "month"]}
+            primaryValueName="Transactions"
+            timeRange={rangeInterval}
+            primaryValueTooltipDescription="txns"
```
acharb, line 87:
```suggestion
            primaryValueTooltipDescription="stroops"
```

frontend/src/components/AverageTransactionFee/index.tsx
```diff
@@ -0,0 +1,94 @@
...
+  const [rangeInterval, setRangeInterval] = useState(TimeRange.DAY);
+
+  useEffect(() => {
+    dispatch(fetchAverageTransactionsFeeData());
+  }, [dispatch]);
+
+  const data = useMemo(() => {
+    if (feeStats.fees) {
+      const formattedMonthlyFee = feeStats.fees.month.map((fee) => ({
+        date: new Date(fee.date),
```
acharb, line 40: let's put the dates in `<day>/<month> <time>` format (so just switching the day and month order from the current format)

and for both the data in the chart and the x-axis format

Screen Shot 2022-07-19 at 3 38 07 PM

frontend/src/components/XML/index.tsx
```diff
@@ -0,0 +1,18 @@
+import { AmountInfoCard } from "components/AmountInfoCard";
+import { SectionCard } from "components/SectionCard";
+
+import "./styles.scss";
+
+export const XML = () => {
+  return (
+    <SectionCard title="XML">
+      <div className="XMLContainer">
+        <AmountInfoCard title="Average Transaction Fee" amount="0.0002 XML" />
```
acharb, line 10: `0.0002 XML` shouldn't be hardcoded. Let's get the average over the last 30 days, from the backend `/api/fees/stats` endpoint

frontend/src/components/FeeStats/index.tsx
```diff
@@ -0,0 +1,115 @@
...
+  const getFeeStats = useCallback(() => {
+    dispatch(fetchFeeStatsDataAction());
+  }, [dispatch]);
+
+  useEffect(() => {
+    getFeeStats();
+    setInterval(() => getFeeStats(), 5 * 1000);
+  }, [getFeeStats]);
+
+  const iconStyle = (value: number) => {
```
acharb, line 48: idk if I agree with these arrow icons. The arrows make it seem like it's a value going up or down (like a trend), when in actually we're using them here just to show magnitude (ie anything over 500 is a downward red arrow)

Screen Shot 2022-07-19 at 3 48 46 PM

maybe something to have our UX team look at after this is merged
cc @jcx120

frontend/src/ducks/feeStats.ts
```diff
@@ -0,0 +1,120 @@
...
+
+export const fetchFeeStatsDataAction = createAsyncThunk<
+  FeeStatsData,
+  void,
+  { rejectValue: RejectMessage; state: RootState }
+>("feeStats/fetchFeeStatsDataAction", async (_, { rejectWithValue }) => {
+  try {
+    const server = new StellarSdk.Server(networkConfig[Network.MAINNET].url);
+
+    const response = server.feeStats();
```
acharb, line 24: idk why typescript allows this but there should be an await here

```suggestion
    const response = await server.feeStats();
```

acharb:
also looks like one of the data values for the Average Transaction Fee data is NaN:

Screen Shot 2022-07-19 at 3 52 45 PM

frontend/src/components/AverageTransactionFee/index.tsx
lccoronel: Yes, we changed this format but in another PR, that's why it didn't reflect, when we merge we believe that all graphs will have the same date format, but I will change here too.

Selhar: Hey @acharb, we weren't clear on what this change meant, in this picture the format is already `DD/MM H:MM`, so we don't really know if it's a data issue on the backend or a display issue change.

lccoronel: @acharb I didn't understand either, I think that x-axis dates are wrong so I fixed this, does make sense now?

Screen Shot 2022-07-20 at 11 34 09

acharb: oops sorry! I meant `<month>/<day>` format

like how the transactions per second chart is

Screen Shot 2022-07-20 at 9 21 18 AM

frontend/src/components/XML/index.tsx
acharb: no let's hardcode those two, according to the [ticket](https://github.com/stellar/dashboard/issues/273)

acharb:
fyi looks like the chart isn't loading in the latest preview

Screen Shot 2022-07-20 at 9 23 41 AM

Selhar:
Was this loading previously? I believe we don't have a cache for fees yet on the preview environment's redis.

acharb:
> Was this loading previously? I believe we don't have a cache for fees yet on the preview environment's redis. For context, the preview environment is unable to make bigquery requests, so it relies exclusively on redis.

ah right, sorry you can ignore. I can add that later today

acharb:
added to redis

acharb (approved):

frontend/src/components/FeeStats/index.tsx
```diff
@@ -0,0 +1,115 @@
...
+import { fetchFeeStatsDataAction } from "ducks/feeStats";
+import "./styles.scss";
+
+export const FeeStats = () => {
+  const { feeStats } = useRedux("feeStats");
+  const dispatch = useDispatch();
+
+  const labels = [
+    { id: "metric", label: "Metric" },
+    { id: "value", label: "value" },
```
quietbits, line 18: Is the label `value` intentionally lower-case?

frontend/src/components/FeeStats/index.tsx
quietbits: I agree with Alec on this one. We should also add a hover label on these to make it more obvious what it's showing. Right now when the cursor hovers over there is nothing.

frontend/src/components/FeeStats/index.tsx
```diff
@@ -0,0 +1,115 @@
...
+    (data: { id: string; name: string }) => {
+      let feeResult;
+      const isCapacityUsage = data.id === "ledger_capacity_usage";
+      const value = get(feeStats.data, data.id);
+
+      if (isCapacityUsage) {
+        feeResult = `${Math.round(value * 100)}%`;
+      } else {
+        const formattedValue = new BigNumber(value).toFormat();
+        feeResult = formattedValue;
```
quietbits, line 86: Maybe we don't need the extra variable here :thinking: 

```suggestion
        feeResult = new BigNumber(value).toFormat();
```

frontend/src/ducks/feeStats.ts
```diff
@@ -0,0 +1,126 @@
...
+  FeeStatsInitialState,
+  Network,
+  RejectMessage,
+} from "types";
+import { networkConfig } from "constants/settings";
+import { formatBigNumbers } from "helpers/formatBigNumbers";
+
+export const fetchFeeStatsDataAction = createAsyncThunk<
+  FeeStatsData,
+  void,
```
quietbits, line 20: Same comment/question as [here](https://github.com/stellar/dashboard/pull/295#discussion_r926988145).

frontend/src/ducks/feeStats.ts
```diff
@@ -0,0 +1,126 @@
...
+  } catch (error) {
+    return rejectWithValue({
+      errorString: getErrorString(error),
+    });
+  }
+});
+
+export const fetchAverageTransactionsFeeData = createAsyncThunk<
+  AverageTransactionFeeData,
+  void,
```
quietbits, line 38: Same as above

frontend/src/helpers/formatBigNumbers.ts
```diff
@@ -0,0 +1,3 @@
+export function formatBigNumbers(bigNumber: string) {
+  return bigNumber.replace(",", "");
+}
```
quietbits, line 3: Could we move this part `new BigNumber(SOME_AMOUNT).toFormat(2)` into the helper rather than asking it to be the input? It will be easier to use the helper if we don't need to remember the specific format of the input.

We could also move it to `/helpers/formatAmount.ts` where we have other amount formatting helpers.

```suggestion
import BigNumber from "bignumber.js";

export function formatBigNumbers(number: string | number) {
  const bigNumber = new BigNumber(number).toFormat(2);
  return bigNumber.replace(",", "");
}
```

frontend/src/types/index.ts
```diff
@@ -300,3 +301,58 @@ export interface FetchLastOperationsActionResponse {
...
+  month: FeesResponse[];
+  day: FeesResponse[];
+}
+
+export interface FeeStatsInitialState {
+  data: FeeStatsData | null;
+  fees: AverageTransactionFeeData | null;
+  status: ActionStatus | undefined;
+  errorString?: string;
+}
```
quietbits, line 358: We should group all ducks types together to make it easier to view/find them.

frontend/src/helpers/formatBigNumbers.ts
lccoronel: Sure 👍🏽

frontend/src/components/FeeStats/index.tsx
lccoronel: So what do we need to show when the cursor hovers over there?


are we going to do that in this PR?

quietbits: It doesn't need to be done in this PR if it wasn't mentioned in the issue.

jcx120:
@lccoronel 

For the module containing "Average Transaction Fee, Base Operation Fee..." etc:
- Let's change the module header label from "XML" to "Transaction Fee Info"
- The currency label should be "XLM" (not "XML")
- Also, given the unit is XLM (Stellar Lumens), looks like "Average Transaction Fee" is displaying Stroops vs XLM? Let's make sure the unit is consistent with the others. 
- In the "Fee Stats" header label, let's link a popout to this [page](https://developers.stellar.org/docs/glossary/fees/#base-fee) for the word "stroops"

Also, in terms of layout, is there a way to move the "Fee stats" module to be nested under this section and within the same UI container? Example:
FeeModule

lccoronel:
@jcx120 about third item, the correct calcule is:

code

the average transaction fee per day divided by 100 stroops(0.00001 XLM), it’s right ?

jcx120:
@lccoronel Let's put the Stroops link on the "STROOPS" word on the left side (not as a separate link).

jcx120:
@lccoronel The current Average Transaction Fee looks like it needs to be multiplied by 0.00001 to get XLM. It is showing currently stroops but is labeled as XLM.

jcx120:
@lccoronel Looks great. 2 more adjustments for the "Transaction Fee Info:" module:

1. Let's put that "Average Transaction Fee" module with the 24hr|30D graph to be **inside** the "Transaction Fee Info" module, below the last 5 ledgers fee stats module. 
2. Label this module: "Average Transaction Fee by Day (In Stroops):" vs "Average Transaction Fee" so its different than the other one above with the same name.

PR #293, opened 2022-06-28
SDFD-28: feat calculation method tps

lccoronel:
* [Reference for src calculation method](https://www.realtps.net/)

* We also need to deliver the average daily TPS over past 30 days.

* Reduce the ledgers endpoint and sum failed + success transactions

* Divide previous result by the amount of seconds calculated with the formula below

RealTPS uses a simple method of counting the transactions in every block over the time period spanning from one week ago until the present moment, then dividing that total number of transactions by the number of seconds from the beginning of the first block until the end of the last block.

Screen Shot 2022-06-28 at 15 58 21


frontend/src/components/TransactionsPerSecond/index.tsx
```diff
@@ -0,0 +1,80 @@
...
+import { fetchTransactionsHistoryAction } from "ducks/transactions";
+import "./styles.scss";
+import { AmountInfoCard } from "components/AmountInfoCard";
+import { Icon } from "@stellar/design-system";
+
+export const TransactionsPerSecond = () => {
+  const { transactions, ledgers } = useRedux("transactions", "ledgers");
+  const dispatch = useDispatch();
+  const lastLedger = ledgers.lastLedgerRecords[0];
+  const selectedTimeInterval = LedgerTransactionHistoryFilterType["30D"];
```
acharb, line 25: I think something needs to change in types/index.ts, since the "30D" endpoint is now `month`, right? 
eg. https://github.com/stellar/dashboard/blob/08acf2a9f7397e881d73bfb45dd2a49b43b22688/backend/src/routes.ts#L33

when I tried running this I got an error that the 30D endpoint doesnt exist

frontend/src/components/TransactionsPerSecond/index.tsx
```diff
@@ -0,0 +1,80 @@
...
+    dispatch(fetchTransactionsHistoryAction(Network.MAINNET));
+  }, [dispatch, selectedTimeInterval]);
+
+  const data = useMemo(() => {
+    const result = transactions.transactionsHistory.items.map((item) => ({
+      date: new Date(item.date),
+      primaryValue: new BigNumber(
+        new BigNumber(item.txTransactionCount / 84600).toFormat(2),
+      ).toNumber(),
+      secondaryValue: 0,
```
acharb, line 37: I'm not very familiar with `VerticalBarChart` but why do we need `secondaryValue` here? Since we're using `primaryValueOnly`, is it needed?

frontend/src/components/TransactionsPerSecond/index.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+  useEffect(() => {
+    dispatch(fetchTransactionsHistoryAction(Network.MAINNET));
+  }, [dispatch, selectedTimeInterval]);
+
+  const data = useMemo(() => {
+    const result = transactions.transactionsHistory.items.map((item) => ({
+      date: new Date(item.date),
+      primaryValue: new BigNumber(
+        new BigNumber(item.txTransactionCount / 84600).toFormat(2),
```
acharb, line 35: this is also dividing the current day by `84600` which will skew results, since the current full day hasn't passed (unless you load this page right at the end of the day)

how about instead we use the `end` data that's returned from the backend to get the total time that's passed?

We'll have to also add the `end` data that's returned from the backend in the duck/transactions.ts

frontend/src/ducks/transactions.ts
```diff
@@ -0,0 +1,91 @@
...
+    try {
+      const historyFilter = ledgerTransactionHistoryConfig["30D"];
+
+      const ledgersUrlRequest = `/api/ledgers${historyFilter.endpointPrefix}${networkConfig[network].ledgerTransactionsHistorySuffix}`;
+      const response = await fetch(ledgersUrlRequest);
+
+      const { data: ledgers } = await response.json();
+
+      return {
+        items: ledgers.map((ledger: Record<string, unknown>) => ({
```
acharb, line 34: let's include the `end` data here too (see my other comment for more details why)

frontend/src/ducks/transactions.ts
```diff
@@ -0,0 +1,91 @@
...
+  Network,
+  RejectMessage,
+  ActionStatus,
+  FetchTransactionsHistoryActionResponse,
+} from "types";
+import { RootState } from "config/store";
+import { getErrorString } from "helpers/getErrorString";
+import { parseDateFromFormat } from "helpers/parseDateFromFormat";
+
+export const fetchTransactionsHistoryAction = createAsyncThunk<
```
acharb, line 18: this is just grabbing the last 30 day history, and since the backend also includes 24h and 1h I think we should label this as `month` as well.

maybe `fetchTransactionsHistoryMonthAction`?

acharb:
Are there designs for the TPS data? I didn't see any in [figma](https://www.figma.com/file/4z9lsHvbsEoVAtzfj3mNzy/Stellar-Dashboard?node-id=206%3A808)

I had two questions about the designs:

1. can the timeline be in format <month>/<day>/<year>?
2. the `Current TPS` chart currently says "in the last ledger". I think that's wrong, since the tps is not just the last ledger. Instead I think it should say "in the last month" if we're using the last 30 days data or something similar?

frontend/src/components/TransactionsPerSecond/index.tsx
```diff
@@ -0,0 +1,80 @@
...
+    }));
+
+    return result;
+  }, [transactions.transactionsHistory]);
+
+  return (
+    <SectionCard
+      title="Transactions per second"
+      titleLinkLabel="API"
+      titleLink={`/api/ledgers${selectedTimeInterval}/public${Network.MAINNET}`}
```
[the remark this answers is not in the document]
lccoronel: I did wrong 😑 

the right way is:

`/api/ledgers${ledgerTransactionHistoryConfig[selectedTimeInterval].endpointPrefix}${networkConfig[Network.MAINNET].ledgerTransactionsHistorySuffix}`

frontend/src/components/TransactionsPerSecond/index.tsx
lccoronel: we don’t need this field, he was like this because before this was a required field

frontend/src/ducks/transactions.ts
lccoronel: of course 👍🏽

lccoronel:
> Are there designs for the TPS data? I didn't see any in [figma](https://www.figma.com/file/4z9lsHvbsEoVAtzfj3mNzy/Stellar-Dashboard?node-id=206%3A808)
> 
> I had two questions about the designs:
> 
> 1. can the timeline be in format `<month>/<day>/<year>`?
> 2. the `Current TPS` chart currently says "in the last ledger". I think that's wrong, since the tps is not just the last ledger. Instead I think it should say "in the last month" if we're using the last 30 days data or something similar?

There aren’t Figma for the TPS data.

1. Yes
2. I will check it

italomlp:
> Are there designs for the TPS data? I didn't see any in [figma](https://www.figma.com/file/4z9lsHvbsEoVAtzfj3mNzy/Stellar-Dashboard?node-id=206%3A808)
> 
> I had two questions about the designs:
> 
> 1. can the timeline be in format `<month>/<day>/<year>`?
> 2. the `Current TPS` chart currently says "in the last ledger". I think that's wrong, since the tps is not just the last ledger. Instead I think it should say "in the last month" if we're using the last 30 days data or something similar?

@acharb We were discussing and trying to handle the differences between the values for the TPS graph and Current TPS. For the graph, we think we fixed it by calculating the closed times as you said before, using the `start` and `end` values from the ledger endpoint data.

For the Current TPS, we were taking a look into the [source code](https://github.com/Aimeedeer/realtps) of [RealTPS](https://www.realtps.net/) website, and we found that [they calculate the TPS by using the `operation_count` value instead of doing the sum `failed_transaction_count + successful_transaction_count`](https://github.com/Aimeedeer/realtps/blob/4f3b67211852c6579f69e4ae5ca5e28e28b37c4d/src/realtps_import/src/clients/stellar.rs#L59) (see function `get_block` from line 46 through line 70). Also, accordingly to its code, it seems that they use the only last ledger to calculate the TPS to show on the website. The close time is calculated by getting the last ledger, and the one previous to it, similar to what we do here.

Our questions are:
1. Should we use `operation_count` to calculate the TPS as they do? Or keep the transactions sum?
2. Should we keep the "Current TPS" as pointing to the last ledger only? Or make something like an average for the last week or last month?

cc @jcx120 @lccoronel @izzyandrade

acharb:
@italomlp let me ask about using ops vs transactions

with regards to the ledger used, it looks like to me (i haven't written rust before so I could be wrong), that they're using the timestamp from a week ago. See here: https://github.com/Aimeedeer/realtps/blob/4f3b67211852c6579f69e4ae5ca5e28e28b37c4d/src/realtps_import/src/calculate.rs#L27

they take the last ledger, and then subtract a weeks worth of seconds from it, and use that as the "previous" ledger

acharb:
@italomlp 
> 1. Should we use operation_count to calculate the TPS as they do? Or keep the transactions sum?

let's use the operation_count like how realTps is doing it

jcx120:
> > Are there designs for the TPS data? I didn't see any in [figma](https://www.figma.com/file/4z9lsHvbsEoVAtzfj3mNzy/Stellar-Dashboard?node-id=206%3A808)
> > I had two questions about the designs:
> > 
> > 1. can the timeline be in format `<month>/<day>/<year>`?
> > 2. the `Current TPS` chart currently says "in the last ledger". I think that's wrong, since the tps is not just the last ledger. Instead I think it should say "in the last month" if we're using the last 30 days data or something similar?
> 
> @acharb We were discussing and trying to handle the differences between the values for the TPS graph and Current TPS. For the graph, we think we fixed it by calculating the closed times as you said before, using the `start` and `end` values from the ledger endpoint data.
> 
> For the Current TPS, we were taking a look into the [source code](https://github.com/Aimeedeer/realtps) of [RealTPS](https://www.realtps.net/) website, and we found that [they calculate the TPS by using the `operation_count` value instead of doing the sum `failed_transaction_count + successful_transaction_count`](https://github.com/Aimeedeer/realtps/blob/4f3b67211852c6579f69e4ae5ca5e28e28b37c4d/src/realtps_import/src/clients/stellar.rs#L59) (see function `get_block` from line 46 through line 70). Also, accordingly to its code, it seems that they use only the last ledger to calculate the TPS to show on the website. The close time is calculated by getting the last ledger, and the one previous to it, similar to what we do here.
> 
> Our questions are:
> 
> 1. Should we use `operation_count` to calculate the TPS as they do? Or keep the transactions sum?
> 2. Should we keep the "Current TPS" as pointing to the last ledger only? Or make something like an average for the last week or last month?
> 
> cc @jcx120 @lccoronel @izzyandrade

Agree with acharb - let's use operation_count and keep it consistent with the real_tps calculation. 
There is no figma for this but I believe UX has reviewed it and looks fine. thanks

jcx120:
Data looks good!

frontend/src/components/TransactionsPerSecond/index.tsx
```diff
@@ -0,0 +1,102 @@
...
+export const TransactionsPerSecond = () => {
+  const { transactions } = useRedux("transactions");
+  const dispatch = useDispatch();
+  const selectedTimeInterval = LedgerTransactionHistoryFilterType["30D"];
+
+  useEffect(() => {
+    dispatch(fetchTransactionsHistoryMonthAction(Network.MAINNET));
+  }, [dispatch]);
+
+  const data = useMemo(() => {
```
acharb, line 35: if you hover over the verticalbar chart entries the dates go backwards. I think it may be because data is sorted in the wrong direction?

Screen Shot 2022-07-11 at 10 56 49 AM

acharb (commented):
nice, this is looking great thx for making those changes.

noticed 1 last bug we should fix before merging

acharb (approved):

Issue #274, opened 2022-05-24
Transactions Module


jcx120:
Build a "Transactions" container component with the following elements:

1. Transactions Per Second (TPS): 120
   - Current TPS
   - [Reference for src calculation method](https://www.realtps.net/) 
2. Historical TPS - last 30 days (Nice to have)
   - Graph chart of average daily TPS over past 30 days, MainNet only
3. Recent Operations 
   - See legacy Dashboard
   - Last 10
   - Fields: Source Operation | Details | Time ago

Considerations:
- The "Transaction" container component will need to be created
- Use [Stellar Design System](https://design-system.stellar.org/) (SDS) components as much as possible.
- Icons should also be used from the SDS ( component). It's OK if they differ slightly.
- Backend data will vary depending on the data element
- On Mainnet and TestNet

raphaelmartins1993:
@jcx120 
Where we can get these operations? Horizon or Hubble?
If Hubble, which tables?

raphaelmartins1993:
hello @jcx120, in the historical TPS of the transactions module we should always bring the last 30 days fixed, or should we also have the possibility of filters such as 1 hour, 1 day, 30 days?
thanks!

jcx120:
@raphaelmartins1993 If this is easy to do, then let's have the 1hr, 1day, 30 day filter to keep it consistent with other metrics. Otherwise, let's keep it fixed at 30 days.

lccoronel:
Hi @jcx120 and @quietbits, we just did the transactions component with "Transactions Per Second" and "Historical TPS", what do you think of the layout?

<img width="839" alt="Screen Shot 2022-06-28 at 15 58 21" src="https://user-images.githubusercontent.com/54275445/176266134-b77fb127-d070-486d-97cc-9bf91f049062.png">

jcx120:
I think it looks fine. We can have Charles review it once a staging instance is available so it can be seen in context.

lccoronel:
ok, we have a preview here: https://dashboard-pr293.prototypes.kube001.services.stellar-ops.com/?network=mainnet

lccoronel:
@jcx120 and @quietbits, another thing, I noticed that the "RECENT OPERATIONS: LIVE NETWORK" table on dashboard v1 some data is treated that is not returning in the [horizon API ](https://horizon.stellar.org/operations?order=desc&limit=200) like: 

* operation name
* operation shares_received
* shares

like others fields.

[...]

PR #294, opened 2022-07-01
SDFD-29: feat recent operations table

lccoronel:
### Recent Operations

* See legacy Dashboard

* Last 10

* Fields: Source Operation | Details | Time ago

Details:
  This is an existing module in the current [Dashboard v1](https://dashboard.stellar.org/)
  The API: https://horizon.stellar.org/operations?order=desc&limit=10
  (We can limit it to 10)

Screen Shot 2022-07-01 at 18 08 35


frontend/src/components/RecentOperations/OperationTypeColumn.tsx
```diff
@@ -0,0 +1,248 @@
...
+
+export const OperationTypeColumn: React.FC<OperationTypeColumnProps> = ({
+  operation,
+  horizonURL,
+}) => {
+  const amount = (
+    am: number,
+    asset_type: string,
+    asset_code: string,
+    asset_issuer: string,
```
quietbits, line 22: With so many params, it might be better to use an object here. We prefer to use camelCased names for params where possible (backend params could be the exception).

```suggestion
{
  amount,
  assetType,
  assetCode,
  assetIssuer,
}: FormatAmountProps
```

frontend/src/components/RecentOperations/OperationTypeColumn.tsx
```diff
@@ -0,0 +1,248 @@
...
+
+    return (
+      <span>
+        {amount}{" "}
+        <AssetLink horizonURL={horizonURL} code={code} issuer={asset_issuer} />
+      </span>
+    );
+  };
+
+  if (operation.type === "create_account") {
```
quietbits, line 42: Why not use switch case for types? It might be easier to read it.

frontend/src/components/RecentOperations/index.tsx
```diff
@@ -0,0 +1,95 @@
+import React, { useCallback, useEffect } from "react";
+import { useDispatch } from "react-redux";
+import { Table } from "@stellar/design-system";
+
+import { useRedux } from "hooks/useRedux";
+import { networkConfig } from "constants/settings";
+import { SectionCard } from "components/SectionCard";
+import { AccountBadge } from "components/AccountBadge";
+import { OperationTypeColumn } from "./OperationTypeColumn";
```
quietbits, line 9: NIT: Relative imports should be at the end (before `./styles.scss`).

frontend/src/components/RecentOperations/index.tsx
```diff
@@ -0,0 +1,95 @@
...
+  useEffect(() => {
+    getLastOperations();
+    setInterval(getLastOperations, 10 * 1000);
+  }, [dispatch, getLastOperations]);
+
+  const renderOperations = (operation: FetchLastOperationsActionResponse) => {
+    const operationLink = `${horizonURL}/operations/${operation.id}`;
+
+    return (
+      <React.Fragment key={Math.random()}>
```
quietbits, line 52: Let's not use `Math.random()` for the key here for better performance. [React docs](https://reactjs.org/docs/reconciliation.html) say this:

> Keys should be stable, predictable, and unique. Unstable keys (like those produced by Math.random()) will cause many component instances and DOM nodes to be unnecessarily recreated, which can cause performance degradation and lost state in child components.

frontend/src/components/RecentOperations/index.tsx
```diff
@@ -0,0 +1,95 @@
...
+  const horizonURL = "https://horizon.stellar.org";
+
+  const labels = [
+    {
+      id: "source",
+      label: "Source",
+    },
+    {
+      id: "operations",
+      label: "Operations",
```
quietbits, line 27: Let's change lthe abel to `Operation`

frontend/src/components/RecentOperations/index.tsx
```diff
@@ -0,0 +1,95 @@
...
+            className="RecentOperations__link"
+            href={operationLink}
+            target="_blank"
+            rel="noreferrer"
+          >
+            {operation.type}
+          </a>
+        </td>
+        <td>
+          <OperationTypeColumn horizonURL={horizonURL} operation={operation} />
```
quietbits, line 71: If possible, could we truncate (add `...`) details if it doesn't fit on one line? We could use CSS for that. This could be an edge case, but it will be nice to ensure it won't break the UI.

frontend/src/constants/knownAccounts.ts
```diff
@@ -0,0 +1,20 @@
+export const knownAccounts = {
+  GCKX3XVTPVNFXQWLQCIBZX6OOPOIUT7FOAZVNOFCNEIXEZFRFSPNZKZT: "Coins base",
```
acharb, line 2: where are these list of accounts and names from?

frontend/src/types/index.ts
```diff
@@ -219,3 +220,82 @@ export interface TransactionsInitialState {
...
   errorString?: string;
 }
+
+export interface OperationsInitialState {
+  lastOperations: FetchLastOperationsActionResponse[];
+  status: ActionStatus | undefined;
+  errorString?: string;
+}
+
+export interface OperationsResponse {
```
acharb, line 230: is there a reason we're not grabbing this type from the stellar js sdk? The response type should def be in there

frontend/src/components/AccountBadge.tsx
```diff
@@ -0,0 +1,61 @@
...
+  id,
+  account,
+}) => {
+  const [isObjectAccount, setIsObjectAccount] = useState({} as ObjectAccount);
+
+  const convertedId = id as unknown as keyof typeof knownAccounts;
+
+  if (
+    knownAccounts[convertedId] ===
+    "GDCIUCGL7VEMMF6VYJOW75KQ5ZCLHAQBRM6EPFTKCRWUYVUOOYQCKC5A"
```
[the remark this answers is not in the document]
acharb: cc @lccoronel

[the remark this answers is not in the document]
lccoronel: @acharb we are try to understand [here](https://github.com/stellar/dashboard/issues/274#issuecomment-1176197284)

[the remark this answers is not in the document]
acharb: let's remove this for now. If we find we need it later we can add it back, it doesn't look like it's critical though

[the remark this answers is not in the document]
acharb: make sure to also remove the account here
https://github.com/stellar/dashboard/blob/8b2db05bb9bdab02558acb3465a5394a8160b476/frontend/src/constants/knownAccounts.ts#L15

frontend/src/types/index.ts
lccoronel: now I fixed it to call the SDK data, but the response is not typed

frontend/src/components/RecentOperations/OperationTypeColumn.tsx
```diff
@@ -0,0 +1,203 @@
...
+  assetType: string;
+  assetCode?: string;
+  assetIssuer?: string;
+}
+
+export const OperationTypeColumn: React.FC<OperationTypeColumnProps> = ({
+  operation,
+  horizonURL,
+}) => {
+  const formatAccount = ({
```
acharb, line 25: this should be formatAmount no?
```suggestion
  const formatAmount = ({
```

acharb (approved):
had suggested 1 var name change otherwise lgtm!

Issue #275, opened 2022-05-24
Ledger module


jcx120:
Build a "Ledger" container component with the following elements:

1. Operations Capacity: Monthly graph of Operations (million) (y axis) vs Time (x axis) with a static max red line indicating capacity limit (Operations (million) per month) 
   - Data is in Hubble
2. % Operations Capacity Utilized current month
   - This is a current month view of utilization 

Considerations:
- The "Ledger" container component will need to be created
- Use [Stellar Design System](https://design-system.stellar.org/) (SDS) components as much as possible.
- Icons should also be used from the SDS ( component). It's OK if they differ slightly.
- Backend data be from Hubble
- On Mainnet (and not TestNet)

raphaelmartins1993:
@jcx120 
Could you bring us more information about [this link](https://stellarfoundation.slack.com/archives/C02B04RMK/p1651588338318189?thread_ts=1634587865.341700&cid=C02B04RMK)? We don’t have permission to access it.

We want to know the maximum capacity to display the red line in the chart.

- This number is static or dynamic?

If you can explain more about what information we need to use in Hubble.

jcx120:
For this module, the query is in Hubble. For details we can reach out to @bartekn 
The link is just the original slack thread that has the context (but not the query)

raphaelmartins1993:
Hi @jcx120. We have a question about the **trendline** that is displayed in the graph. Since we haven't seen any specification about it, we would like to know if it should be shown or not?
If it should be displayed, can you tell us if it is calculated or static?
Thank you very much!

jcx120:
That trendline is not MVP scope and can be left out (I think its a r^2 trendline generated in excel). 

We do want that Max red line which is a constant for the maximum # of operations per month based on theoretical capacity of 5000 operations per second. @bartekn can you share that number? thanks

jcx120:
I like option 3 but once its prototyped we can also have our UX designer provide input. thanks

italomlp:
Oh, sorry, maybe I was not so clear. They are the same component, in light and dark mode variants. Also, the one that has a tooltip (image 2) is just the light mode variant with the mouse hovering over the chart. Does that make sense?

bartekn:
[Here's](https://github.com/stellar/dashboard/issues/275#issuecomment-1147795658) how you can calculate it.

lccoronel:
@acharb and @jcx120, in the graph of the ledgers module, do you need to have a period variation?

[...]

PR #295, opened 2022-07-12
SDFD-32: feat ledger module

lccoronel:
##Ref for Operations Capacity: [Ledger module · Issue #275 · stellar/dashboard](https://github.com/stellar/dashboard/issues/275#issuecomment-1147795658) 

### Considerations:

* Trendline is not necessary to implement;

* The number of red line It's 518,400,000.

* The "Ledger" container component will need to be created

* Use [Stellar Design System](https://design-system.stellar.org/) (SDS) components as much as possible.

* Icons should also be used from the SDS ( component). It's OK if they differ slightly.

* Retrieve data for calculations from SDK

* On Mainnet (and not TestNet)

Screen Shot 2022-07-25 at 09 27 40

### DIFF
--- a/frontend/src/types/index.ts
+++ b/frontend/src/types/index.ts
@@ -367,9 +367,3 @@ export interface AverageTransactionFeeData {
   month: FeesResponse[];
   day: FeesResponse[];
 }
-
-export interface FetchLedgerModuleResponse {
-  date: Date;
-  primaryValue: number;
-}
-


frontend/src/ducks/ledgers.ts
```diff
@@ -141,6 +143,32 @@ export const fetchLedgersTransactionsHistoryAction = createAsyncThunk<
   },
 );
 
+export const fetchLedgersModule = createAsyncThunk<
```
acharb, line 146: lets make the name more general, eg. `fetchLedgerOperations`

frontend/src/components/LedgerModule/index.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+import "./styles.scss";
+
+import { fetchLedgersModule } from "ducks/ledgers";
+import { SectionCard } from "components/SectionCard";
+import { VerticalBarChart } from "components/charts/VerticalBarChart";
+import { useRedux } from "hooks/useRedux";
+import { ActionStatus } from "types";
+
+const limitLine = 518400000;
```
acharb, line 12: let's add a comment here for where this number is coming from
eg

```
// max operations per month = 12*1000*60*24*30 = 518400000
```

frontend/package.json
```diff
@@ -100,6 +100,7 @@
     "html-webpack-plugin": "^5.5.0",
     "mini-css-extract-plugin": "^2.6.0",
     "prettier": "^2.5.1",
+    "sass": "^1.53.0",
```
acharb, line 103: just curious what's this needed for?

frontend/src/components/LedgerModule/index.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+  useEffect(() => {
+    dispatch(fetchLedgersModule());
+  }, [dispatch]);
+
+  return (
+    <SectionCard
+      title="Ledger Module"
+      titleLinkLabel="API"
+      titleLink={`/api/ledgers`}
```
acharb, line 26: this titleLink is incorrect, should be `/api/ledgers/hour/public` I think

frontend/src/ducks/ledgers.ts
```diff
@@ -141,6 +143,32 @@ export const fetchLedgersTransactionsHistoryAction = createAsyncThunk<
...
+    const response = await fetch(
+      `/api/ledgers${historyFilter.endpointPrefix}${
+        networkConfig[Network.MAINNET].ledgerTransactionsHistorySuffix
+      }`,
+    );
+
+    const operations = await response.json();
+
+    const result = operations.data.map((operation: LedgerModuleItem) => ({
+      date: new Date(operation.end),
```
acharb, line 162: we should use a serializable data for this. If you look at the console logs this throws an error:
```
react_devtools_backend.js:4026 A non-serializable value was detected in the state, in the path: `ledgers.ledgerModule.0.date`. Value: Thu Jun 23 2022 09:38:06 GMT-0700 (Pacific Daylight Time) 
```

frontend/package.json
lccoronel: because `node-sass` lib was depreciated and we can’t use on machines with M1 chip, so when use `sass` everything works 😊

frontend/src/ducks/ledgers.ts
```diff
@@ -141,6 +143,35 @@ export const fetchLedgersTransactionsHistoryAction = createAsyncThunk<
...
+    const response = await fetch(
+      `/api/ledgers${historyFilter.endpointPrefix}${
+        networkConfig[Network.MAINNET].ledgerTransactionsHistorySuffix
+      }`,
+    );
+
+    const operations = await response.json();
+
+    const result = operations.data.map((operation: LedgerModuleItem) => ({
+      date: parseDateFromFormat(
```
acharb, line 162: this `fetchLedgerOperations` call is failing with error `Invalid time value`, probably the parsing is incorrect somehow

lccoronel: 😬 sorry I didn't see that

frontend/src/components/LedgerOperations/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+        networkConfig[Network.MAINNET].ledgerTransactionsHistorySuffix
+      }`}
+      isLoading={ledgers.status === ActionStatus.PENDING}
+      noData={!ledgers.ledgerOperations.length}
+    >
+      {ledgers.ledgerOperations.length > 0 && (
+        <div className="LedgerOperations__mainChart">
+          <div className="LedgerOperations__mainChart__container">
+            <VerticalBarChart
+              data={data}
```
acharb, line 50: data dates look backwards, I think we should sort before passing here

frontend/src/components/LedgerOperations/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+import { fetchLedgerOperations } from "ducks/ledgers";
+import { SectionCard } from "components/SectionCard";
+import { VerticalBarChart } from "components/charts/VerticalBarChart";
+import {
+  ledgerTransactionHistoryConfig,
+  networkConfig,
+} from "constants/settings";
+
+// max operations per month = 12*1000*60*24*30 = 518400000
+const limitLine = 518400000;
```
acharb, line 17: actually we should lower this number to the max per day, not per month, since that's what this chart is showing
(@jcx120 correct me if I'm wrong)

so instead it would be
```
// max operations per day = 12*60*24*1000 = 17,280,000
const limitLine = 17280000;
```

jcx120: The Max ops line is per month. The X axis is currently shown in the staging instance as daily but should be monthly per the original example in the [doc](https://docs.google.com/document/d/1J2HruGsFGluNSY6IXJ9ooSnIYW5gv_LM-udYnIkkoaw/edit).

acharb:
hey @lccoronel we actually need the chart to be per month. See Jesse's comment here: https://github.com/stellar/dashboard/pull/295#discussion_r922365375

So the red line should go back to `518400000` (that was changed in the last [commit](https://github.com/stellar/dashboard/pull/295/commits/d493df7738598900585d43d0fc951ba29add8fa4))

And should look similar to (chart in the [docs](https://docs.google.com/document/d/1J2HruGsFGluNSY6IXJ9ooSnIYW5gv_LM-udYnIkkoaw/edit)):

Screen Shot 2022-07-18 at 12 08 20 PM


I think this will probably involve adding a new backend endpoint, as from I know we don't have this data available. So we'll have to grab it from Big Query cc @Selhar

Selhar:
What data do we need? Currently ledgers endpoint returns on a monthly, daily and hourly interval. @acharb

Selhar:
Hey @acharb, just to confirm, this cache will take potentially more than a day to generate, given the amount of data we're bringing in, is this ok? We'll only have to generate it once in production though, but it will take a really long time.

acharb:
@Selhar ya  I think that should be fine 🤔 we could probably run as infrequent as once a month.

What's the query you're running?
(or I can take a look whenever the PR comes out)

Selhar:
Oh it's the same query we already use for current ledgers, the only change is the filter goes to january 1 2018.

acharb:
ah, I think we should just grab the operations/month from big query, instead of "catching up" like we do with the current backend ledger endpoints

frontend/src/components/RecentOperations/index.tsx
```diff
@@ -12,7 +12,11 @@ import { OperationTypeColumn } from "./OperationTypeColumn";
...
 
-export const RecentOperations = () => {
+export const RecentOperations = ({
+  network = Network.MAINNET,
+}: {
+  network?: Network;
+}) => {
   const { operations } = useRedux("operations");
   const dispatch = useDispatch();
   const horizonURL = "https://horizon.stellar.org";
```
quietbits, line 22: Not part of this PR, why do we hard code Horizon URL here instead of getting it from `networkConfig` in `constants/settings.ts`?

frontend/src/ducks/ledgers.ts
```diff
@@ -141,6 +143,32 @@ export const fetchLedgersTransactionsHistoryAction = createAsyncThunk<
   },
 );
 
+export const fetchLedgerOperations = createAsyncThunk<
+  FetchLedgerOperationsResponse[],
+  void,
```
quietbits, line 148: Should we use `undefined` instead of `void` here for the accepted parameter (or not param in this case)? `void` feels more like a type for a function that doesn't return anything. I'm not very opinionated about it, just curious what you think. :slightly_smiling_face:

quietbits (commented):
The website breaks when switching to the Testnet tab. We might want to add `id` check in the `AccountBadge` component to make sure it won't crash the website. It might be good to double-check the data as well, just to make sure it's OK that it doesn't return the `id` value we're expecting.

image

lccoronel:
@acharb @quietbits  @jcx120  

After @Selhar merged the [ledgers operations PR](https://github.com/stellar/dashboard/pull/297), we start to change the ledgers operation chart and it was like this:

Screen Shot 2022-07-22 at 08 57 29

except for the chart bars, the rest of the chart data is not yet integrated with the API, but you can already have a preview, is it as you imagined?

Selhar:
i think this data is in the opposite direction, isn't it? As in, 2022 is to the left, and 2015 to the right.

frontend/src/components/LedgerOperations/index.tsx
```diff
@@ -0,0 +1,64 @@
...
+    <SectionCard
+      title="Ledger Operations"
+      titleLinkLabel="API"
+      titleLink={`/api/ledgers${historyFilter.endpointPrefix}${
+        networkConfig[Network.MAINNET].ledgerTransactionsHistorySuffix
+      }`}
+      isLoading={ledgers.status === ActionStatus.PENDING}
+      noData={!ledgers.ledgerOperations.length}
+    >
+      {ledgers.ledgerOperations.length > 0 && (
```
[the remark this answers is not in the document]
lccoronel: Actually I think that we don’t need this validation more 😁

frontend/src/components/RecentOperations/index.tsx
lccoronel: sorry, I took as base the dashboards v1, I will fix it

lccoronel:
> i think this data is in the opposite direction, isn't it? As in, 2022 is to the left, and 2015 to the right.

Yes, I just put the data on the chart to show how it would look with a lot of data, I haven't organized it yet

frontend/src/ducks/ledgers.ts
lccoronel: I prefer use `undefined`, but I left `void` following the design pattern, in some other functions they were already like this.

quietbits: Let's use `undefined`, we prefer it as well. We must have missed `void` :sweat_smile: . Thank you!

acharb:
besides the backwards data the charts looking good to me

we need to change the "max" line back to `518,400,000` though, since now we're doing per month and not per day

quietbits (approved):

PR #296, opened 2022-07-14
SDFD-36_fee_stats_data

Selhar:
Hey, i wasn't sure if we wanted a new module just for fees, so i left it on dex_data (since we already bring a lot of related datapoints there). Let me know if a separate module/endpoint is ideal @acharb. 

Sample output:
image

I'm pushing this early so @lccoronel can finish the front end adjustements. Meanwhile i'll work on tests and other time intervals.

### DIFF
--- a/backend/src/fees/index.ts
+++ b/backend/src/fees/index.ts
@@ -4,8 +4,8 @@ import { getFeesData1d, getFeesData30d } from "./data";
 export async function getFeeStats(_: any, res: Response, next: NextFunction) {
   try {
     return res.json({
-      day: getFeesData1d(),
-      month: getFeesData30d(),
+      day: await getFeesData1d(),
+      month: await getFeesData30d(),
     });
   } catch (e) {
     next(e);


Selhar:
Sent in the hourly interval, from what i understand for this task we only need monthly and last 24hrs intervals. Let me know if there are any changes or improvements you'd like on this PR @acharb

backend/src/dex/index.ts
```diff
@@ -61,6 +63,10 @@ export async function getAll(_: any, res: Response, next: NextFunction) {
       paymentsLast24h: await getPaymentsData(),
       uniqueAssets: await getUniqueAssetsData(),
       activeAccounts: await getActiveAccountsData(),
+      fees: {
+        month: await getFeesData30d(),
+        hour: await getFeesData1d(),
```
acharb, line 68:
```suggestion
        day: await getFeesData1d(),
```

acharb:
> Hey, i wasn't sure if we wanted a new module just for fees, so i left it on dex_data (since we already bring a lot of related datapoints there).

ya  I think it makes sense to have a new fee module instead of trying to group into DEX data stuff, to make it clear this data isn't directly related to DEX data

I think something like `/api/fees` would make sense

Selhar:
Committed and testes the changes you've made, it seems to be sound here! Let me know if you want any other changes @acharb

acharb (approved):
lgtm!

PR #297, opened 2022-07-21
Get operations stats from ledger history

Selhar:
Adds new endpoint `/api/ledgers/op_stats`, it returns operations sorted by month. Starts on 2015, ends on current month. Sample output:
image

The query is pretty fast actually, takes about 4 seconds. The worry i had for long executing queries was because i thought we wanted a similar system to what we have for other ledgers.

I think when we implement the redis CRON jobs, we can clear out this key as well whenever we want more updated results. The key is `ledger-operation-stats`. Let me know if you want any improvements in this PR @acharb

### DIFF
--- a/backend/src/ledgers/data.ts
+++ b/backend/src/ledgers/data.ts
@@ -262,7 +262,7 @@ export async function getOpStats() {
   const query = `
     SELECT 
       FORMAT_DATE('%Y-%m', cast(closed_at as date)) as closing_date,
-      SUM(operation_count) as operations 
+      SUM(tx_set_operation_count) as operations 
     FROM ${bigQueryEndpointBase}.history_ledgers 
     GROUP BY closing_date 
     ORDER BY closing_date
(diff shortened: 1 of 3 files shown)


acharb:
this lgtm, I'm confirming the BQ query is correct with our team first before approving

backend/src/ledgers/data.ts
```diff
@@ -256,3 +257,18 @@ export async function updateCache(
   );
   await redisClient.set(pagingTokenKey, pagingToken);
 }
+
+export async function getOpStats() {
+  const query = `
+    SELECT 
+      FORMAT_DATE('%Y-%m', cast(closed_at as date)) as closing_date,
+      SUM(operation_count) as operations
```
acharb, line 265: let's use the tx_set_operation_count field, which gives all the operations (operation_count just gives the successful ones)
```suggestion
      SUM(tx_set_operation_count) as operations 
```

Selhar: I tested locally, it checks out!

acharb (approved):
lgtm 👍 

also I added the data to preview redis

PR #310, opened 2022-08-17
add script for refreshing redis to be run by cronjob

acharb:
some of the new v2 endpoints (dex, fees, ledger op stats) have data that updates when redis is cleared, so adding the code to do that that will be triggered by a cron job

### DIFF
--- a/backend/package.json
+++ b/backend/package.json
@@ -11,7 +11,9 @@
     "start": "UPDATE_DATA=true DEV=true ts-node ./src/app.ts",
     "clean": "rm -rf node_modules && rm -rf build",
     "test": "jest --forceExit",
-    "build": "tsc --project ."
+    "build": "tsc --project .",
+    "redis:reset_day": "ts-node ./src/redis/redisRefresh.ts day",
+    "redis:reset_month": "ts-node ./src/redis/redisRefresh.ts month"
   },
   "dependencies": {
     "@google-cloud/bigquery": "^5.11.0",
--- a/backend/package.json
+++ b/backend/package.json
@@ -12,8 +12,8 @@
     "clean": "rm -rf node_modules && rm -rf build",
     "test": "jest --forceExit",
     "build": "tsc --project .",
-    "redis:reset_day": "ts-node ./src/redis/redisRefresh.ts day",
-    "redis:reset_month": "ts-node ./src/redis/redisRefresh.ts month"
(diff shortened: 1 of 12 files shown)


backend/src/dex/data.ts
```diff
@@ -71,11 +82,35 @@ export async function getUniqueAssetsData() {
...
+  const activeAccounts = await fetchCachedData(
+    REDIS_DEX_KEYS.DEX_ACTIVE_ACCOUNTS,
+    query,
+  );
+
+  return activeAccounts[0]?.data;
+}
+
+//TODO: Currently unused due to a pending update to how we fetch asset prices. Once they're available we should refactor and make use of the function below again.
+const cache24hKey = "dex-volume-sum-24h";
```
acharb, line 110: I moved this stuff to the bottom of the file under this TODO comment, confused me we're not using this (yet)

backend/src/dex/data.ts
```diff
@@ -1,14 +1,19 @@
...
 import {
   bigQueryEndpointBase,
   fetchBigQueryData,
   fetchCachedData,
 } from "../utils";
 
-const cache24hKey = "dex-volume-sum-24h";
-const cache48hKey = "dex-volume-sum-48h";
-const cacheOverallKey = "dex-volume-sum-overall";
+export const REDIS_DEX_KEYS = {
```
acharb, line 9: also moved the redis keys to constants that weren't already

backend/package.json
```diff
@@ -11,7 +11,9 @@
     "start": "UPDATE_DATA=true DEV=true ts-node ./src/app.ts",
     "clean": "rm -rf node_modules && rm -rf build",
     "test": "jest --forceExit",
-    "build": "tsc --project ."
+    "build": "tsc --project .",
+    "redis:reset_day": "ts-node ./src/redis/redisRefresh.ts day",
```
acharb, line 15: on second look I think the trigger will have to be part of the Dockerfile 🤔  , but I'll change that  in a second PR if needed

piyalbasu: I have limited context, but my hunch is that you'll want to trigger this from the Dockerfile

piyalbasu (approved):

acharb:
@piyalbasu fyi added this small commit https://github.com/stellar/dashboard/pull/310/commits/19b5115efe8ec4eca997b17d70a6b83db1ff06cd
the production command will be ran from Docker/kube but DEV=true is needed when running it locally from package.json

PR #317, opened 2022-09-14
link to op_stats in total monthly operations chart

acharb:
when verifying the v2 data I noticed one of the API links was to the wrong backend endpoint, so changed that here.

Also, the Total Monthly Operations chart (tracks sum of operations since 2015) we only have mainnet data for, so hiding that chart if we're on testnet

### DIFF
--- a/frontend/src/components/TotalMonthlyOperations/index.tsx
+++ b/frontend/src/components/TotalMonthlyOperations/index.tsx
@@ -63,7 +63,7 @@ export const TotalMonthlyOperations = ({
             </div>
           </div>
 
-          <RecentOperations network={network} noShadow={true} />
+          <RecentOperations network={network} />
         </SectionCard>
       ) : (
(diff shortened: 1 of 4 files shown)


frontend/src/components/TotalMonthlyOperations/index.tsx
```diff
@@ -46,30 +40,34 @@ export const TotalMonthlyOperations = ({
...
-            maxLineOffset={550000000}
-          />
-        </div>
-      </div>
+    <>
+      {network === Network.MAINNET ? (
+        <SectionCard
+          title="Total monthly operations"
+          titleLinkLabel="API"
+          titleLink={`/api/ledgers/op_stats`}
```
acharb, line 48: link change here, was pointing to `/api/ledgers/month/public` (if on mainnet)

quietbits (approved):
:rocket:

PR #330, opened 2023-11-03
Switch Ubuntu base image to 22.04 from 20.04

satyamz:
### What

Switch Ubuntu to 22.04 from 20.04
### Why

We are moving other images to 22.04, this change improves consistency across SDF docker images.


jacekn (approved):
TLS cert problem is known.
We'll also check if this repo can be decomissioned under https://github.com/stellar/ops/issues/2711

PR #345, opened 2026-01-09
Migrate to Vite + other updates

quietbits:
- Used AI for most of the updates to save time.
- Migrated from
  - Gulp + Webpack 1 to Vite v4,
  - Node v16 to v22,
  - React v16 to v18.
- Made as few updates as possible to get this working, keeping the same UI.
- We're using the `muicss` framework for the UI, and it hasn't been [updated in 5 years](https://www.npmjs.com/package/muicss). Because the whole UI depends on it, we can't update most of the other dependencies.
- Created custom D3 charts because the `react-d3-components` library doesn't work with newer Node versions; [it's no longer maintained](https://www.npmjs.com/package/react-d3-components).
- If we want to keep maintaining this project, we should do a proper update with a complete rewrite, because in this state, there are too many outdated parts to update them properly.
- [PR Preview](https://dashboard-pr346.dashboard-previews.kube001.services.stellar-ops.com/) running the current version for comparison.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,19 +4,19 @@
 
 To build this project, you must have the following dependencies installed:
 
-- node 10.16.3
-- yarn
+- Node v22
+- NPM
 
 ## Installation
 
 ```sh
-yarn
+npm install
 ```
 
 ## Developing
 
 ```sh
-yarn start
+npm run dev
 ```
 
 ### If you wish to use backend server API, you need to have redis running locally on port 6379 (default for redis)
@@ -27,7 +27,7 @@ yarn start
 brew install redis
 ```
 
-(Other install directions can be found here: https://redis.io/download)
+(Other install directions can be found [here](https://redis.io/download))
 
 Make sure it's running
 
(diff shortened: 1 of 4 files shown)


frontend/components/PublicNetworkLedgersHistoryChart.jsx
```diff
@@ -1,16 +1,30 @@
 import React from "react";
 import Panel from "muicss/lib/react/panel";
 import axios from "axios";
-import { scale, format } from "d3";
-import BarChart from "react-d3-components/lib/BarChart";
+import * as d3 from "d3";
+import D3BarChartNoXLabels from "./D3BarChartNoXLabels.jsx";
 import clone from "lodash/clone";
```
jeesunikim: this is not being used

common/lumens.js
```diff
@@ -47,85 +47,74 @@ const accounts = {
...
+  return response.data._embedded.records[0].total_coins;
 }
 
-exports.inflationLumens = inflationLumens;
-function inflationLumens() {
-  return Promise.all([
+export async function inflationLumens() {
+  const [totalLumensValue, originalSupply] = await Promise.all([
     totalLumens(horizonLiveURL),
     ORIGINAL_SUPPLY_AMOUNT,
```
jeesunikim: dear cluade, this is hardcoded, no reason to use `Promise.all`

frontend/components/AccountBalance.jsx
```diff
@@ -1,5 +1,5 @@
 import React from "react";
-import AmountWidget from "./AmountWidget";
+import AmountWidget from "./AmountWidget.jsx";
```
jeesunikim: It should be just `./AmountWidget`. There are five different files that use `/AmountWidget.jsx`. It should be corrected to just `/AmountWidget`

frontend/components/App.jsx
```diff
@@ -3,24 +3,25 @@ import Panel from "muicss/lib/react/panel";
...
-import Incidents from "./Incidents";
-import LedgerCloseChart from "./LedgerCloseChart";
-import LumensCirculating from "./LumensCirculating";
-import LumensNonCirculating from "./LumensNonCirculating";
-import PublicNetworkLedgersHistoryChart from "./PublicNetworkLedgersHistoryChart";
-import RecentOperations from "./RecentOperations";
-import TotalCoins from "./TotalCoins";
-import TransactionsChart from "./TransactionsChart";
-import FailedTransactionsChart from "./FailedTransactionsChart";
+import AppBar from "./AppBar.jsx";
```
jeesunikim: Those import files don't need `.jsx`. But we can leave them for now since they aren't critical. I am just flagging.

jeesunikim (approved):

